PR #51, opened 2017-07-14
Add sso login form

### LINKED ISSUE #50: Add SSO login form
Add single sign on login form, where e.g. dashboard can redirect it's users to to log in to the system.

dennyverbeeck:
This adds an SSO login form for OAauth2 based SSO. Closes #50 .


dennyverbeeck:
Indeed, I forgot about /api/account, UserResource is now reverted back to the original.

The OAuth2 login and authorization forms are rendered by the Spring MVC itself (as opposed to having an angular app that talks to the spring app like the rest of ManagementPortal's UI). This way we can directly use the OAuth2 functionalities of Spring Security.

PR #56, opened 2017-09-01
Add config to dev profile to let hibernate update the database schema.

dennyverbeeck:
Otherwise we need to add the sequences we reference in @SequenceGenerator annotations explicitly to Liquibase changelog. For dev environments this solution should be fine.


dennyverbeeck:
This is already fixed in another branch that will be merged into dev, no need for this PR.

PR #59, opened 2017-09-05
Add oauth client lib

### LINKED ISSUE #25: Invenstigate how RADAR-Rest-Api can be integrated to use authorization with MP
We need to investigate the best way to integrate Rest-Api to use security mechanisms of MP to secure REST endpoints in rest-api application.
One cohesive option would be convert Rest-api as a spring project and use it as a `ResourceServer` and use MP's `AuthorizationServer` to ensure the authorization.  (Preferable) 
Otherwise we could provide a small distribution with annotations and integrate it with Rest-api.

dennyverbeeck:
Add library for managing a token in client apps using the client credentials grant.

Closes #25

### DIFF
--- a/oauth-client-util/src/main/java/org/radarcns/oauth/OAuth2AccessToken.java
+++ b/oauth-client-util/src/main/java/org/radarcns/oauth/OAuth2AccessToken.java
@@ -1,0 +1,198 @@
+package org.radarcns.oauth;
+
+/*
+ * Copyright 2017 King's College London
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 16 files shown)


oauth-client-util/src/main/java/org/radarcns/oauth/OAuth2Client.java
```diff
@@ -0,0 +1,131 @@
...
+ * Class for handling OAuth2 client credentials grant
+ */
+public class OAuth2Client {
+    private URL tokenEndpoint;
+    private String clientId;
+    private String clientSecret;
+    private Set<String> scope;
+    private OAuth2AccessToken currentToken;
+
+    private static OkHttpClient HTTP_CLIENT;
```
nivemaham: I understand we have to have to have an `OAuth2Client` for each client credentials. In that case, would it still make sense to have a static `HTTP_CLIENT` that authorizes using credentials from instance variables?

oauth-client-util/src/main/java/org/radarcns/oauth/OAuth2AccessToken.java
```diff
@@ -0,0 +1,191 @@
...
+
+    @JsonProperty("error")
+    private final String error;
+
+    @JsonProperty("error_description")
+    private final String errorDescription;
+
+    @JsonProperty("message")
+    private final String message;
+
```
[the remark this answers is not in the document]
dennyverbeeck: Yeah this class is a bit confusing. It actually captures the response from ManagementPortal, rather than being a token. So i renamed it to OAuth2AccessTokenDetails. I also added documentation to clarify the intent of this class. The JWT token is contained in the `accessToken` field. I purposely did not include a JWT parser in this library. A client of this library might not need it (essentially you only need to provide the token to a protected resource), but if they do need it, they are free to choose a JWT library, they might already be using one. So information like `authorities`, if needed by the client, can be parsed from the JWT.

oauth-client-util/src/main/java/org/radarcns/oauth/OAuth2Client.java
dennyverbeeck: Initially i did not have it static. But thanks to @fnobilia for pointing out that `OkHttpClient` is designed to be used this way. It has its own thread and connection pool. So the recommended use is to have only one instance of the class. Reference [here](https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.html).

PR #60, opened 2017-09-05
Remove JDBCTokenStore

dennyverbeeck:
We are using JWT tokens, a JWTTokenStore is defined. We don't need to define a JDBCTokenStore as well. In fact the jdbc token store interferes with proper JWT token operation if they're defined simultaneously.


dennyverbeeck:
We remove the JDBC token store, because we don't need to check a JWT access token with a central repository of tokens. JWT contains a signature that can be verified by anyone. Since there is no JDBC token store anymore, we also can not remove or revoke a token for a removed user. This means that the user will be able to continue using his token for the remainder of the expiration time, which a known 'downside' of using JWT tokens.
However, in the traditional OAuth setting with a central token store, protected resources often cache the result of a token lookup. This means that users will be able to use their token for the remainder of the cache time (unless an explicit cache invalidation mechanism is implemented).
So since the 'downside' of JWT, i.e. not being able to revoke tokens, also would occur in the setting with a JDBC token store because of the cache, there is no reason anymore to keep using it in combination with our JWT tokens.

PR #61, opened 2017-09-05
Redcap integration

nivemaham:
1. Allow meta-data for subject and project
2. Add an additional Authority for `ROLE_EXTERNAL_ERF_INTEGRATOR` with limited resource access
3. Create a UUID as a subject.login to uniquely identify a `Subject in the system
4. Allow to discontinue a subject. 
5. Additional `API` to query a subject based on `projectId` and `externalId`.


src/main/java/org/radarcns/management/web/rest/SubjectResource.java
```diff
@@ -101,9 +119,25 @@
...
-       log.debug("REST request to get all Subjects");
-       return subjectService.findAll();
+    public ResponseEntity getAllSubjects(
+        @RequestParam(value = "projectId" , required = false) Long projectId,
+        @RequestParam(value = "externalId" , required = false) String externalId) {
+        log.error("ProjectID {} and external {}" , projectId, externalId);
+        if(projectId!=null && externalId!=null) {
+            Subject subject = subjectRepository.findOneByProjectIdAndExternalId(projectId, externalId).get();
+            SubjectDTO subjectDTO = subjectMapper.subjectToSubjectDTO(subject);
+            return ResponseUtil.wrapOrNotFound(Optional.ofNullable(subjectDTO));
```
dennyverbeeck: I think we should return a `List<SubjectDTO>` here as well. This way consumers of the API know what the response type is and know it will always be the same. Otherwise a client would always need to test for list or single entity. We can create a list from a single element easily with e.g. `Collections.singletonList(subjectDTO)`. Then you can also make the type signature more specific to `ResponseEntity<List<SubjectDTO>>`.

src/main/webapp/i18n/en/global.json
```diff
@@ -90,6 +90,16 @@
...
+        },
+        "attribute" : {
+            "title" : "Attributes",
+            "key" : "Attribute Key",
+            "value" : "Attribute Value",
+            "add" : "Add",
+            "error" : {
+                "alreadyExist" : "This attribute key already exist in the list"
+            }
+
```
dennyverbeeck: Can placeholders be added to the `nl` version of this and the other JSONs for the added properties? This way we keep them in sync.

PR #66, opened 2017-09-18
Multiple projects with new project view

nivemaham:
A new user interface to support multiple projects for a user.
1. A user gets started with the new `User-management` view, where a user can be created with one or more roles. A role is defined by the combination of an authority and a project. 
2. A `Project-view` where a user could view project details, and project specific entities such as sources, subjects, device types etc.
3. A `General-view` limited for `SYS_ADMIN` who can view all entities regardless of projects and modify the entities if necessary.


src/main/webapp/app/entities/project/project-detail.component.html
```diff
@@ -1,59 +1,113 @@
+<div id="wrapper" [ngClass]="{'toggled': true}">
+    <div id="sidebar-wrapper">
+        <div class="sidebar-nav" *ngIf="project">
+            <p class="status-header">
+                {{project.projectName}}
+            </p>
+            <ul class="status-list">
+                <ng-container>
+                    <li [ngClass]="{'status-group-selected': showSubjects , 'status-group': !showSubjects}"
```
dennyverbeeck: I think we should add a `cursor: pointer` CSS to these list items so it is more obvious to the user they are clickable

src/main/java/org/radarcns/management/repository/SubjectRepository.java
```diff
@@ -18,10 +19,13 @@
...
 
-    @Query("select distinct subject from Subject subject left join fetch subject.sources where subject.user.project.id = :projectId")
+    @Query("select distinct subject from Subject subject left join fetch subject.sources "
+        + "left join fetch subject.user user "
+        + "join user.roles roles where roles.project.id = :projectId")
     List<Subject> findAllByProjectId(@Param("projectId") Long projectId);
 
-    @Query("select subject from Subject subject left join fetch subject.sources where subject.id =:id")
+    @Query("select subject from Subject subject left join fetch subject.sources "
+        + "where subject.id =:id")
```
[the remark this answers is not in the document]
nivemaham: I think it is not necessary since `subject` has a `oneToOne` relationship between `user`. It should be automatically handled by the framework. In the above query it was necessary to have it explicitly. We are using this method for `api/subjects/{id}` and it is retrieving necessary data. So i think it is not necessary to have it written. WDYT?

src/main/webapp/app/entities/project/project-detail.component.html
nivemaham: Good point. Done!

PR #67, opened 2017-09-19
Updates to datamodel

dennyverbeeck:
- Add device_version field to DeviceType. A new version of a device type might appear, we need to capture this.
- Add multicolumn unique index on DeviceType fields producer, model and version
- Add source_name field to Source. The idea is that a source_id can be a unique, custom identifier (e.g. a UUID), and the source_name is for display purposes, but also needs to be unique.


src/main/webapp/app/shared/source/source-dialog.component.html
```diff
@@ -13,6 +13,18 @@ <h4 class="modal-title" id="myDeviceLabel" jhiTranslate="managementPortalApp.sou
...
+                   required />
+            <div [hidden]="!(editForm.controls.sourceName?.dirty && editForm.controls.sourceName?.invalid)">
+                <small class="form-text text-danger"
+                       [hidden]="!editForm.controls.sourceName?.errors?.required" jhiTranslate="entity.validation.required">
+                    This field is required.
+                </small>
+            </div>
+        </div>
+        <div class="form-group">
             <label class="form-control-label" jhiTranslate="managementPortalApp.source.sourceId" for="field_devicePhysicalId">Source Id</label>
```
nivemaham: i think a user cannot enter a `source-id` during creation/edit according to our latest model. So this field should be removed. 
```
<div class="form-group">
            <label class="form-control-label" jhiTranslate="managementPortalApp.source.sourceId" for="field_devicePhysicalId">Source Id</label>
            <input type="text" class="form-control" name="sourceId" id="field_devicePhysicalId"
                [(ngModel)]="source.sourceId"
            required />
            <div [hidden]="!(editForm.controls.sourceId?.dirty && editForm.controls.sourceId?.invalid)">
                <small class="form-text text-danger"
                   [hidden]="!editForm.controls.sourceId?.errors?.required" jhiTranslate="entity.validation.required">
                   This field is required.
                </small>
            </div>
        </div>
```

src/main/java/org/radarcns/management/service/dto/SourceDTO.java
```diff
@@ -16,6 +16,9 @@
     @NotNull
```
nivemaham: As we agreed a UUID should be assigned to the `sourceId` automatically by the system upon creation. This should be either handled either in `Domain` object using hibernate or in DTO's setter. Are we missing this?

dennyverbeeck: Indeed, it missed it in cherry-picking the commits from my working branch to this branch, it will come shortly in the next PR!

dennyverbeeck:
Thanks, very useful! I have incorporated your comments in my working branch for the source registration feature. So i will go ahead with this merge now.

PR #68, opened 2017-09-19
Source registration merged

### LINKED ISSUE #31: Dynamic device registration for mobile device registration
As mentioned in [issue#20](https://github.com/RADAR-CNS/ManagementPortal/issues/20), we want to allow dynamic device registration with dynamic ids for mobile devices.
The model should be enhanced to have  a hasDynamicId for a deviceType and should provide a rest-resource to register a device. 
The registered device should be listed under device list as the other devices.

### LINKED ISSUE #40: API for querying participant device assignments
Request for comments from @fnobilia, @nivemaham, @dennyverbeeck.

In the pRMT app, the user ID and assigned source IDs can be figured out from the JWT. The app needs a few calls to consolidate its device assignment:

1. what device types are assigned in the project where a user is participant
2. what device types the devices that are already assigned have. The device types should at least include the Android sensor provider (plugin) that should be used to read from the device, and possibly also frequencies and custom properties (like battery-dependent frequencies or activity-dependent frequencies).
3. register source IDs for devices that are not yet registered.

The first two pieces of information could be provided in a single request:
```
HTTP GET /api/user/devices
```
Response:
```json
HTTP 200
{
  "devices": [
    {
      "id": "some-uuid",
      "expectedName": "A00017",
      "expectedSourceId": "null",
      "deviceType": {
        "id": "my-uuid",
        "name": "EmpaticaE4",
        "topics": {
          "android_empatica_e4_acceleration": {
            "key_schema": "org.radarcns.key.MeasurementKey",
            "value_schema": "org.radarcns.empatica.EmpaticaE4Acceleration",
            "provider": "org.radarcns.empatica.EmpaticaE4Provider",
            "enabled": true
          }
        }
      }
    },
    {
      "id": "some-uuid2",
      "sourceId": "abcd-ewaefafa-agagagbawe-asdfas",
      "deviceType": {
        "id": "my-uuid2",
        "name": "AndroidPhone",
        "topics": {
          "android_phone_acceleration": {
            "key_schema": "org.radarcns.key.MeasurementKey",
            "value_schema": "org.radarcns.phone.PhoneAcceleration",
            "frequency": 16,
            "provider": "org.radarcns.phone.PhoneSensorProvider",
            "enabled": true
          },
          "android_phone_light": {
            "key_schema": "org.radarcns.key.MeasurementKey",
            "value_schema": "org.radarcns.phone.PhoneLight",
            "frequency": 16,
            "provider": "org.radarcns.phone.PhoneSensorProvider",
            "enabled": false
          },
          "android_phone_call": {
            "key_schema": "org.radarcns.key.MeasurementKey",
            "value_schema": "org.radarcns.phone.PhoneCall",
            "send_interval": 43200,
            "provider": "org.radarcns.empatica.PhoneCallProvider",
            "enabled": true
          }
        }
      }
    }
  ]
}
```

If a `dynamic` device type was registered and assigned to a project, when this request hits, probably an empty device with that device type should be created. If `device.expectedName` and `device.expectedSourceId` are `null`, then any device ID might be registered. The reasoning is, that a clinician may not know the technical source ID, but only the name of a device (e.g. "Empatica E4 - A00017". In another case, the source ID is autogenerated by the plugin (e.g. phone sensors) so the clinician cannot know the source ID in advance.

New devices could be registered as follows:

```json
HTTP POST /api/user/devices
{
  "devices": [
    {
      "id": "some-uuid",
      "name": "Empatica E4 - A00017",
      "sourceId": "AB:01:CD:23:EF:45",
      "firmware": "E4-2.1.0",
      "model": "E4.A",
      "os": null
    }
  ]
}
```
Response:
```
HTTP 204
```
or for an illegal request,
```json
HTTP 400
{
   "message": "Name or ID does not match acceptedIds."
}
```

The request uses the same UUIDs as provided in the previous response, and it registers a fixed `sourceId`, plus possibly some optional metadata. The backend should check that either the name or the MAC-address should match what was in `acceptableIds`. 

After registration of all source IDs, the app should request a new JWT, which will include those source IDs.

dennyverbeeck:
This branch merged the `source_registration` branch with `dev` in prepration of this PR. This PR implements the dynamic source registration feature for clients. A user with the correct permissions can now `POST /api/subjects/{id}/sources` a list of sources. Closes #31. Closes #40.


src/main/java/org/radarcns/management/web/rest/SubjectResource.java
```diff
@@ -211,4 +228,74 @@ else if( projectId!=null) {
...
+     * in the source object.
+     *
+     * @param sourceDTOS List of sources to assign
+     * @return The updated Subject information
+     */
+    @PostMapping("/subjects/{id}/sources")
+    @Timed
+    @Secured({AuthoritiesConstants.SYS_ADMIN, AuthoritiesConstants.PROJECT_ADMIN,
+              AuthoritiesConstants.PARTICIPANT})
+    public ResponseEntity<List<SourceDTO>> assignSources(@PathVariable Long id,
```
nivemaham: I prefer `POST subjects/{login}/sources` for external clients, this would keep everything based on the `subject-id/login` for external partners. We don't need to additionally share the persistent layer id

src/main/java/org/radarcns/management/service/dto/SourceDTO.java
```diff
@@ -13,8 +13,7 @@
 
     private Long id;
 
-    @NotNull
-    private String sourceId;
+    private UUID sourceId;
```
nivemaham: if we gonna have `UUID` in SourceDTO which we need to change the typescript to receive `UUID` type. The `Source id` text box should be removed when from `general.subject-dialog.component.html` as we agreed that we will not enter a `UUID` manually.

src/main/java/org/radarcns/management/service/mapper/ProjectMapper.java
```diff
@@ -28,6 +29,8 @@
 
     MinimalProjectDetailsDTO projectToMinimalProjectDetailsDTO(Project project);
 
+    MinimalProjectDetailsDTO projectDTOToMinimalProjectDetailsDTO(ProjectDTO projectDTO);
```
nivemaham: It is a bit unconventional  to map one DTO to another DTO in `Mapstruct`. Why do we need this?

src/main/java/org/radarcns/management/web/rest/DeviceTypeResource.java
```diff
@@ -52,6 +62,16 @@ public DeviceTypeResource(DeviceTypeService deviceTypeService) {
         if (deviceTypeDTO.getId() != null) {
             return ResponseEntity.badRequest().headers(HeaderUtil.createFailureAlert(ENTITY_NAME, "idexists", "A new deviceType cannot already have an ID")).body(null);
         }
+        DeviceTypeDTO existing = deviceTypeService.findByProducerAndModel(
+            deviceTypeDTO.getDeviceProducer(), deviceTypeDTO.getDeviceModel());
```
nivemaham: as mentioned earlier, are we missing the `version` in the query ?

src/main/java/org/radarcns/management/web/rest/DeviceTypeResource.java
```diff
@@ -52,6 +62,16 @@ public DeviceTypeResource(DeviceTypeService deviceTypeService) {
...
+        if (existing != null) {
+            Map<String, String> errorParams = new HashMap<>();
+            errorParams.put("message", "A DeviceType with the specified producer and model "
+                + "already exists. This combination needs to be unique.");
+            errorParams.put("producer", deviceTypeDTO.getDeviceProducer());
+            errorParams.put("model", deviceTypeDTO.getDeviceModel());
```
nivemaham: Wouldn't it be better if we send a `HttpResponse` than an exception? How would the client be noticed in case of the exception?

src/main/java/org/radarcns/management/security/ClaimsTokenEnhancer.java
```diff
@@ -60,13 +61,20 @@ public OAuth2AccessToken enhance(OAuth2AccessToken accessToken,
                             + role.getAuthority().getName())
                         .collect(Collectors.toList());
                     additionalInfo.put("roles", roles);
+
+                }
+
+                Subject subject = subjectRepository.findBySubjectLogin(userName);
```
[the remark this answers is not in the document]
dennyverbeeck: Yes the `userName` is part of the JWT, but the `subject-id` is it's primary key. The `UserResource` endpoints are based on `login`, but the `SubjectResource` endpoints are based on `id`. So either we add the `subject-id`, or we change the endpoints in `SubjectResource` to be based on `login`. I chose not to modify the endpoints, but either solution is fine for me.

src/main/java/org/radarcns/management/repository/ProjectRepository.java
```diff
@@ -19,6 +20,9 @@
     @Query("select project from Project project left join fetch project.deviceTypes where project.id =:id")
     Project findOneWithEagerRelationships(@Param("id") Long id);
 
+    @Query("select project from Project project left join fetch project.deviceTypes where project.projectName =:name")
+    Project findOneWithEagerRelationshipsByName(@Param("name") String name);
```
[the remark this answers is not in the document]
dennyverbeeck: Indeed, good point!

src/main/java/org/radarcns/management/repository/DeviceTypeRepository.java
```diff
@@ -20,4 +20,16 @@
     @Query("select deviceType from DeviceType deviceType left join fetch deviceType.sensorData where deviceType.id =:id")
     DeviceType findOneWithEagerRelationships(@Param("id") Long id);
 
+    @Query("select deviceType from DeviceType deviceType left join fetch deviceType.sensorData "
+        + "where deviceType.deviceProducer =:producer and deviceType.deviceModel =:model")
+    DeviceType findOneWithEagerRelationshipsByProducerAndModel(
```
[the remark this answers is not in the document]
dennyverbeeck: Yes `deviceVersion` should indeed be part of this as well.

src/main/java/org/radarcns/management/service/SourceService.java
```diff
@@ -60,9 +70,53 @@
...
+            if (source.getProject().getId() != null) {
+                project = projectService.findOne(source.getProject().getId());
+            } else if (source.getProject().getProjectName() != null) {
+                project = projectService.findOneByName(source.getProject().getProjectName());
+            }
+            source.setProject(projectMapper.projectDTOToProject(project));
+        }
+        // generate defaults for source id and source name if they were not provided
+        UUID uuid = UUID.randomUUID();
+        if (source.getSourceName() == null) {
```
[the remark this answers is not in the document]
dennyverbeeck: The `@NotNull` is checked only on endpoints where `@Valid` annotation is added to the `@RequestBody`. So in the case of `POST /subjects/{id}/sources` it is possible to provide a `SourceDTO` without `sourceName`, in which case a default value is generated.
That being said, i don't think there's a reason this behavior should not occor on `POST /sources`, i think we should remove the `@NotNull` on `sourceName`.

src/main/java/org/radarcns/management/web/rest/SubjectResource.java
dennyverbeeck: Indeed, i prefer it as well

src/main/java/org/radarcns/management/security/ClaimsTokenEnhancer.java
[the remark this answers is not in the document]
nivemaham: True. But as this endpoint is specific for client registration and will not be a part of MP front-end interaction, i think we can go for `subject/{login}sources` for this specific api. so that the external parties will not get confused between `subject.getId` and `subject-id` which is the login. It would keep things simpler IMO

src/main/java/org/radarcns/management/service/dto/SourceDTO.java
dennyverbeeck: Do you mean from `general.source-dialog.component.html` instead? It is removed there already, or rather, it will be shown only, and read-only, when it is already part of the model, i.e. when editing an existing source.
The serialization and de-serialization of the DTO takes care of converting from and to the appropriate string representation of a `UUID` type.

src/main/java/org/radarcns/management/service/mapper/ProjectMapper.java
dennyverbeeck: It seems to be unused, possibly a leftover from an earlier implementation, agreed it should be removed.

src/main/java/org/radarcns/management/web/rest/DeviceTypeResource.java
dennyverbeeck: The `ExceptionTranslator` takes care of transforming a raised exception to a proper HTTP response, with appropriate code. This is also the class that responds with `InternalServerError` by default for exceptions. By providing an implementation for a specific exception in `ExceptionTranslator` we can inform the client more clearly about why the request did not process successfully.

src/main/java/org/radarcns/management/service/dto/SourceDTO.java
nivemaham: indeed. it was not the `subject`. But what i meant was the `app/shared/sources/source-dialog-component.html`.  But it is not very user-friendly to enter an UUID manually. Isn't it

src/main/java/org/radarcns/management/service/DeviceTypeService.java
```diff
@@ -39,6 +44,8 @@
     public DeviceTypeDTO save(DeviceTypeDTO deviceTypeDTO) {
         log.debug("Request to save DeviceType : {}", deviceTypeDTO);
         DeviceType deviceType = deviceTypeMapper.deviceTypeDTOToDeviceType(deviceTypeDTO);
+        List<SensorData> savedSensorData = sensorDataRepository.save(deviceType.getSensorData());
+        deviceType.setSensorData(new HashSet<>(savedSensorData));
         deviceType = deviceTypeRepository.save(deviceType);
```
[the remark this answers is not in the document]
nivemaham: Apparently, this doesn't quite work for `ManyToMany` relationship. I have reverted the changes for now and kept what you have implemented. Will investigate on this later.

src/main/webapp/app/entities/device-type/device-type.model.ts
```diff
@@ -13,7 +13,7 @@ export class DeviceType {
         public sourceType?: SourceType,
         public sensorDataId?: number,
         public projectId?: number,
-        public hasDynamicId?: number,
+        public canRegisterDynamically?: number,
```
dennyverbeeck: Shouldn't this be a `boolean`?

PR #72, opened 2017-09-28
Make primary keys of demo data negative

dennyverbeeck:
The database primary key sequences are initialized at 0. Having demo data with positive primary keys will cause the first few insert operations to fail since the primary keys already exist.


dennyverbeeck:
That's also a possibility, and `sequenceGenerator` is supposed to be more efficient in hibernate compared to `GenerationType.IDENTITY`. Ok, reverted the PK changes and added the `sequenceGenerator`s.

PR #74, opened 2017-09-28
Radar auth library

dennyverbeeck:
Library for components of the RADAR platform to validate tokens, and a single place to store the permissions logic. The permissions are a static permission matrix right now, in the future this can be expanded, or even replaced with a more powerful mechanism such as ACLs. But for now this should suffice for our use cases.


dennyverbeeck:
Indeed, i'm currently reworking the MP security rules to use these `Permission` objects and `RadarAuthorization` class. Basically, spring security will check **authentication** (i.e. unauthenticated users can not access the rest endpoints), and this library will do **authorization** (i.e. is the authenticated user allowed to perform the requested action).

nivemaham:
Shouldn't it be the other way around ?

PR #79, opened 2017-10-02
new source assign/remove component

nivemaham:
new source assign/remove component. Closes [58](https://github.com/RADAR-CNS/ManagementPortal/issues/58)

### DIFF
--- a/src/main/java/org/radarcns/management/service/SourceService.java
+++ b/src/main/java/org/radarcns/management/service/SourceService.java
@@ -169,10 +169,40 @@ public class SourceService {
         sourceRepository.delete(id);
     }
 
+    /**
+     * Returns all sources by project in {@link SourceDTO} format
+     * @param projectId
+     * @return list of sources
+     */
     public List<SourceDTO> findAllByProjectId(Long projectId) {
         return sourceRepository.findAllSourcesByProjectId(projectId)
             .stream()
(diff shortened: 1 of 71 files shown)


src/main/java/org/radarcns/management/service/SourceService.java
```diff
@@ -169,10 +169,40 @@ public void delete(Long id) {
...
+    /**
+     * Returns all sources by project in {@link MinimalSourceDetailsDTO} format
+     * @param projectId
+     * @return list of sources
+     */
+    public List<MinimalSourceDetailsDTO> findAllMinimalSourceDetailsByProject(Long projectId) {
+        return sourceRepository.findAllSourcesByProjectId(projectId)
+            .stream()
+            .map(sourceMapper::sourceToMinimalSourceDetailsDTO)
+            .collect(Collectors.toCollection(LinkedList::new));
```
strangepleasures: Is there any reason to use LinkedList everywhere? It's only good for doing insertions and deletions and even then it's probably better to use LinkedHashSet. Why not just `collect(Collectors.toList())`?

nivemaham: @strangepleasures cannot think of a reason why jhipster chose `LinkedList` over a `List`.  As we have similar pattern in all services, we will take this a separate improvement issue.

Issue #83, opened 2017-10-02
Provide deviceType, sourceId and expectedSourceName in /api/subjects/{login}


blootsvoets:
To know what devices should be registered and what devices are already registered, the pRMT app needs the following information from all sources listed in `/api/subjects/{login}`:
- `deviceType`: the ID listed under `project.deviceTypes`
- `sourceId`: the source ID that the pRMT is to use for that source
- `expectedSourceName`: an optional custom string to do custom matching with a device.

At least for pRMT, the numeric ID and `deviceTypeAndSourceId` are not used.

PR #84, opened 2017-10-03
Changed Minimal Source DTO

blootsvoets:
Added the properties described in #83, plus any properties needed by the MP UI.

### DIFF
--- a/src/main/java/org/radarcns/management/service/dto/MinimalSourceDetailsDTO.java
+++ b/src/main/java/org/radarcns/management/service/dto/MinimalSourceDetailsDTO.java
@@ -1,14 +1,17 @@
 package org.radarcns.management.service.dto;
 
+import java.util.UUID;
+
 /**
  * Created by nivethika on 13-6-17.
  */
 public class MinimalSourceDetailsDTO {
-
     private Long id;
-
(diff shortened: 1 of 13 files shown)


src/main/java/org/radarcns/management/service/dto/MinimalSourceDetailsDTO.java
```diff
@@ -1,14 +1,17 @@
...
 /**
  * Created by nivethika on 13-6-17.
  */
 public class MinimalSourceDetailsDTO {
-
     private Long id;
-
-    private String deviceTypeAndSourceId;
-
+    private Long deviceType;
```
nivemaham: May be we could call this `deviceTypeId`

src/main/webapp/i18n/nl/source.json
```diff
@@ -22,7 +22,7 @@
             "unassigned": "Unassigned",
             "deviceType": "Device Type",
             "project": "Project",
-            "expectedSourceName" : "Expected source name",
+            "expectedSourceName" : "Physical ID",
```
nivemaham: This change is not reflected on `en` version at ` src/main/webapp/i18n/en/source.json`

PR #85, opened 2017-10-04
Some consistency fixes for api/subjects/{login}/sources

blootsvoets:


src/main/java/org/radarcns/management/service/mapper/SourceMapper.java
```diff
@@ -39,11 +39,13 @@
 
     @Mapping(source = "sourceId" , target = "sourceId")
     @Mapping(source = "expectedSourceName" , target = "expectedSourceName")
+    @Mapping(source = "deviceType.id" , target = "deviceTypeId")
     @Mapping(source = "deviceType.deviceModel" , target = "deviceTypeModel")
     @Mapping(source = "deviceType.deviceProducer" , target = "deviceTypeProducer")
     @Mapping(source = "deviceType.catalogVersion" , target = "deviceCatalogVersion")
     @Mapping(source = "project.id" , target = "projectId")
     @Mapping(source = "assigned" , target = "assigned")
```
nivemaham: Mapping of same attributes will work with or without Mapping annotation.

blootsvoets:
Yes, good point. I'll take a stab at that.

src/main/java/org/radarcns/management/service/mapper/SourceMapper.java
blootsvoets: Thanks, I wasn't aware

src/main/java/org/radarcns/management/service/dto/SubjectDTO.java
```diff
@@ -44,7 +46,7 @@
 
     private Set<MinimalSourceDetailsDTO> sources = new HashSet<>();
 
-    private Set<AttributeMapDTO> attributes = new HashSet<>();
+    private Map<String, String> attributes = new HashMap<>();
```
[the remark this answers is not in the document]
blootsvoets: Ok, created a Javascript `DictionaryMapperComponent` to use a map of `[key: string]: string` instead. This also solves some uniqueness and interaction bugs.

nivemaham:
looks good. I think later on we can convert projectDTO also to use Map directly. Good solution @blootsvoets

PR #95, opened 2017-10-18
Pair app show relevant oauth clients

### LINKED ISSUE #88: Pair-app should only contain relevant oauth-clients
Paid-app currently loads all clients with `authorization-code` flow of the oauth-clients for QR-code generation. We should find a mechanism to load the relevant clients only (e.g. without managementportal-app)

dennyverbeeck:
Now a specific entry in the additional_information map is necessary to have an OAuth client show up in the pair app dialog. Updated the Readme as well with this information.

Closes #88

### DIFF
--- a/src/main/java/org/radarcns/management/service/mapper/ClientDetailsMapperImpl.java
+++ b/src/main/java/org/radarcns/management/service/mapper/ClientDetailsMapperImpl.java
@@ -10,6 +10,7 @@ import org.springframework.stereotype.Component;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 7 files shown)


src/main/java/org/radarcns/management/service/mapper/ClientDetailsMapperImpl.java
```diff
@@ -52,6 +55,8 @@ public ClientDetails clientDetailsDTOToClientDetails(ClientDetailsDTO detailsDTO
         if (detailsDTO.getAutoApprove()) {
             result.setAutoApproveScopes(Collections.unmodifiableSet(detailsDTO.getScope()));
         }
+        result.setAdditionalInformation(Collections.unmodifiableMap(detailsDTO
```
strangepleasures: It seems that detailsDTO.getAdditionalInformation() can return null. In that case Collections.unmodifiableMap will produce a NPE. Same for other fields properties.

dennyverbeeck: Good point, added the necessary null checks in this class.

src/main/java/org/radarcns/management/service/mapper/ClientDetailsMapperImpl.java
```diff
@@ -43,31 +52,41 @@ public ClientDetails clientDetailsDTOToClientDetails(ClientDetailsDTO detailsDTO
...
-        result.setAdditionalInformation(Collections.unmodifiableMap(detailsDTO
-            .getAdditionalInformation()));
+        result.setAdditionalInformation(detailsDTO.getAdditionalInformation());
         return result;
     }
 
     @Override
     public List<ClientDetailsDTO> clientDetailsToClientDetailsDTO(List<ClientDetails> detailsList) {
+        if (Objects.isNull(detailsList)) {
+            return null;
```
strangepleasures: It's better to return Collections.emptyList() here, otherwise you would have to check for null everywhere it's used.

src/main/java/org/radarcns/management/service/mapper/ClientDetailsMapperImpl.java
```diff
@@ -43,31 +52,41 @@ public ClientDetails clientDetailsDTOToClientDetails(ClientDetailsDTO detailsDTO
...
+        }
         return detailsList.stream().map(this::clientDetailsToClientDetailsDTO)
             .collect(Collectors.toList());
     }
 
     @Override
     public List<ClientDetails> clientDetailsDTOToClientDetails(List<ClientDetailsDTO> detailsDTOList) {
+        if (Objects.isNull(detailsDTOList)) {
+            return null;
```
strangepleasures: Same as above

PR #97, opened 2017-10-20
Bring master upto date with dev

yatharthranjan:


dennyverbeeck:
Since we want to follow the Git-flow workflow, a release branch should be created from which to merge. I'll create one for `0.1.1` when #107 is merged in.

blootsvoets:
As an addition: tags should generally be made on the master branch or if this is not possible in a dedicated release branch. Master branch is preferable though...

PR #107, opened 2017-10-26
Remove demo data from prod profile

### LINKED ISSUE #91: Remove demo data for production profile
Remove demo data for production profile and load only `authorities`, `roles`, `roles_user` and `users` with default admin:admin user. The rest should be created by actual users.

dennyverbeeck:
Demo data is now only loaded with the dev and test profiles. The admin user is always loaded.

Closes #91


src/main/resources/config/application-prod.yml
```diff
@@ -22,10 +22,10 @@ spring:
     datasource:
         type: com.zaxxer.hikari.HikariDataSource
         driverClassName: org.postgresql.Driver
-        url: jdbc:postgresql://localhost:5432/managementportal
```
nivemaham: better not change the database name. there are other configurations that need to be changed accordingly.

src/main/resources/config/application-prod.yml
```diff
@@ -22,10 +22,10 @@ spring:
...
+        url: jdbc:postgresql://192.168.56.101:5432/mp
 #        name: managementportal
-        username: <username>
-        password: <password>
+        username: mp
+        password: secret
```
strangepleasures: Is this the actual database password? If so, it's not a good idea to put it on GitHub. If this is just a default value, I would prefer it to be something like CHANGE_ME.

src/main/resources/config/liquibase/admin_user.csv
```diff
@@ -0,0 +1,2 @@
+"ID";"LOGIN";"PASSWORD_HASH";"FIRST_NAME";"LAST_NAME";"EMAIL";"ACTIVATED";"LANG_KEY";"CREATED_BY";"LAST_MODIFIED_BY"
+1;admin;$2a$10$gSAhZrxMllrbgj/kkK9UceBPpChGWJA7SYIb1Mqo.n5aNLq1/oRrC;Administrator;Administrator;admin@localhost;true;en;system;system
```
strangepleasures: Is it safe to put the password hash on GitHub?

src/main/resources/config/application-prod.yml
dennyverbeeck: Indeed, i was testing the production profile locally and accidentally committed the changes. Changes to this file have been reverted.

src/main/resources/config/application-prod.yml
dennyverbeeck: Indeed, i was testing the production profile locally and accidentally committed the changes. Changes to this file have been reverted.

src/main/resources/config/liquibase/admin_user.csv
dennyverbeeck: We need a way for a user to first login, as in most applications this is `admin:admin`. Of course the admin should change the password of their first login.

Issue #105, opened 2017-10-26
Frontend does not refresh token after expiration


dennyverbeeck:
Currently, after the JWT expires the user needs to login again, even if he is still working in the app. The frontend should refresh the JWT after it expires if the user is still working in the app.

nivemaham:
Apparently, this is on purpose for security purposes. If necessary we can implement this. But the jhipster suggests to login again.

dennyverbeeck:
Hmm, IMO it's odd behavior to have to log in again in an application where i'm still active. I would understand the logout after a period of inactivity. But currently the logout is even if the user is active. If this is the desired behavior then fine with me of course, but i suspect users would get annoyed pretty quickly. Let's follow up in one of the calls to see what others think.

afolarin:
I think it would unusual to be logged out while active (I don't think I've come across this behaviour in other systems) and could result in data loss. Perhaps an alternative is the timeout could trigger a dialog that requires the user to confirm they wish to remain logged in?

dennyverbeeck:
Sounds good @afolarin, i'll get on it

nivemaham:
Hi guys, i did a background check on this again and these are the known concerns that are discussed. 
[Should we refresh token](https://github.com/jhipster/generator-jhipster/issues/441)
[session storage or localstorage to store token](https://github.com/jhipster/generator-jhipster/issues/3405 )
and an approach took by jhipster to solve it by using server-side sessions. Our current approach will need some change if we want to do the same. 

My proposal is to use the `Remember me` button when siging-in. If the user choose to use remember-me we store the token in sessionStorage and refresh the token when it expires for now. What do you think? @dennyverbeeck , @blootsvoets

dennyverbeeck:
The `Remember me` button is supposed to be used to remember the user between sessions ([docs](https://docs.spring.io/spring-security/site/docs/3.0.x/reference/remember-me.html)). I think the token should be refreshed when the user is active, regardless of the remember me function. Probably the simplest approach would be to implement something like this in the `auth-oauth2.service.ts` `getToken()` function:

1. If access token is not expired, return it
2. Otherwise, if refresh token is not expired, use it to get new access token, return to 1.
3. If refresh token is expired as well, prompt for new login.

[...]

Issue #92, opened 2017-10-09
Do not hard code client credentials on front-end


nivemaham:
To have secured deployment in production we need to load runtime environment variables. Possible way to do this is via loading an `config.json` file from assets on module `OnInit`. This will eliminate hard coded client-credentials of MP. 

see these [link](https://stackoverflow.com/questions/36528824/pass-environment-variables-to-an-angular2-app

PR #114, opened 2017-11-06
Release 0.2.0

dennyverbeeck:
Key features now include:
- Creation of projects, subjects, devices, etc.
- Updated authentication/authorization and permissions systems
- Source assigner for assigning sources to subjects
- Ability to pair mobile apps to subjects by scanning a QR code from the app
- Updated project view
- Demo data no longer loaded when production profile is selected
- Fix issue that prevents new users from setting up a password
- Fix CORS issue on production profile
- Fix existing browser-based e2e tests
- Add new browser-based e2e tests
- `radar-auth`: Java library for authenticating clients that have tokens issued by ManagementPortal
- `oauth-client-util`: Java library for managing OAuth tokens in a client application using the client credentials grant type


nivemaham:
Should we wait till we fix #105 and #92 before we make a major release?

dennyverbeeck:
This is still a minor release, i think it's beneficial to release more often. I've also already set up an automated Docker build. There are a number of critical fixes in this branch already like the reset password bug and enabling CORS, needed for the development of the aRMT app. Let's release this as soon as possible so other developers can also continue their work, and include the fixes we're currently working on in a next minor or patch release?

yatharthranjan:
@dennyverbeeck agreed. Also we need the new version to be released for starting the enrolment process.

dennyverbeeck:
As per offline discussion, will go ahead with the merge and release.

PR #115, opened 2017-11-07
Exclude /api/account/reset_password endpoints from JwtAuthenticationFilter

### LINKED ISSUE #112: Cannot login with new users created
Cannot login using the users created through the management portal user management page and email activation link.
Also cannot login using users created by requests to the users API.

dennyverbeeck:
This change is necessary for people to be able to set their password for the first time using the set password link emailed to them.

Closes #112

### DIFF
--- a/src/main/java/org/radarcns/management/config/SecurityConfiguration.java
+++ b/src/main/java/org/radarcns/management/config/SecurityConfiguration.java
@@ -107,7 +107,22 @@ public class SecurityConfiguration extends WebSecurityConfigurerAdapter {
     public FilterRegistrationBean jwtAuthenticationFilterRegistration() {
         FilterRegistrationBean registration = new FilterRegistrationBean();
         registration.setFilter(jwtAuthenticationFilter());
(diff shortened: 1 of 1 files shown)


nivemaham:
It makes sense to do this. However, i'm not convinced that this will resolve the #112. Plus i am wondering how did it work without this JWT filtering applied to all these resources before.

nivemaham:
How about `/login/` url? How should it be covered.

dennyverbeeck:
The `JwtAuthenticationFilter` validates the supplied JWT and adds it to the servlet request. Before this change all paths starting with `/api` would be checked. However for `/api/account/reset_password/init` and `/api/account/reset_password/finish` we expect that there will not be a JWT there since these endpoints are called by unauthenticated users. Similarly for `/login`, this is called by unauthenticated users so there will not be a JWT so we do not configure `JwtAuthenticationFilter` for those URLs.
Unfortunately ServletFilters do not have a means to exclude URLs, so now we need to list all `/api` endpoints explicitly, omitting the `reset_password` ones. Hope this makes it more clear!

PR #128, opened 2017-11-13
Merge release 0.2.1 back into dev

dennyverbeeck:
This includes updates to the version numbers and updates to the protractor tests


nivemaham:
Why cannot we include/update the versions of the sub-projects accordingly prior to the release? @dennyverbeeck Is there any limitations?

dennyverbeeck:
Following the git-flow workflow, we update the versions on the release branch. Technically, i should have merged from the release branch into master and from release branch into dev as well when doing the release, which i forgot. I'll remember it next time!

PR #129, opened 2017-11-13
configure zuul proxy and filters , store tokens in cookies

### LINKED ISSUE #117: Token storage and CSRF protection
@nivemaham did research into this and found these articles: [Should we refresh token](https://github.com/jhipster/generator-jhipster/issues/441), [session storage or localstorage to store token](https://github.com/jhipster/generator-jhipster/issues/3405 ) and [an approach took by jhipster to solve it by using server-side sessions](https://github.com/jhipster/generator-jhipster/issues/5752).

We are storing our JWT in web storage which makes it vulnerable to XSS attacks. Recommendation is to store it as an http-only cookie. We also don't have CSRF protection on our api endpoints, we do have it on the OAuth endpoints though.

nivemaham:
1. Configured zuul proxy in between MP front-end and backend.
2. Added filters to remove `refresh-token` from response and store it in cookie
3. Stores `access-token` in cookie. 
4. Removes hard-coded client credentials from MP front-end and make it configurable in back-end. 

Closes #117 , #92

### DIFF
--- a/src/main/java/org/radarcns/management/filters/OAuth2TokenRequestPostZuulFilter.java
+++ b/src/main/java/org/radarcns/management/filters/OAuth2TokenRequestPostZuulFilter.java
@@ -1,0 +1,91 @@
+package org.radarcns.management.filters;
+
+import static org.radarcns.management.filters.OAuth2TokenRequestPreZuulFilter.REFRESH_TOKEN_COOKIE;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Map;
+
+import javax.servlet.http.Cookie;
+
+import org.apache.commons.io.IOUtils;
+import org.radarcns.management.config.ManagementPortalProperties;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.zuul.ZuulFilter;
+import com.netflix.zuul.context.RequestContext;
+
+/**
+ * A post-filter for all the request sent to Zuul proxy.
+ * This investigates the response body and stores the Refresh token in a Cookie and remove it from response body.
+ */
+@Component
+public class OAuth2TokenRequestPostZuulFilter extends ZuulFilter {
+
+    @Autowired
+    private ManagementPortalProperties managementPortalProperties;
+    private final Logger logger = LoggerFactory.getLogger(this.getClass());
+    private final ObjectMapper mapper = new ObjectMapper();
+
+    @Override
+    public Object run() {
+        final RequestContext ctx = RequestContext.getCurrentContext();
+        logger.debug("in zuul filter " + ctx.getRequest().getRequestURI());
+
+        final String requestURI = ctx.getRequest().getRequestURI();
+        final String requestMethod = ctx.getRequest().getMethod();
+
+        try {
+            final InputStream is = ctx.getResponseDataStream();
+            String responseBody = IOUtils.toString(is, "UTF-8");
+            if (responseBody.contains("refresh_token")) {
+                final Map<String, Object> responseMap = mapper
+                    .readValue(responseBody, new TypeReference<Map<String, Object>>() {
+                    });
+                final String refreshToken = responseMap.get("refresh_token").toString();
+                responseMap.remove("refresh_token");
+                responseBody = mapper.writeValueAsString(responseMap);
+
+                final Cookie cookie = new Cookie(REFRESH_TOKEN_COOKIE, refreshToken);
+                cookie.setHttpOnly(true);
+                cookie.setSecure(true);
+                cookie.setPath(ctx.getRequest().getContextPath() + "oauthserver/oauth/token");
+                cookie.setMaxAge(this.managementPortalProperties.getFrontend().getSessionTimeout()); // 30 minites
+                ctx.getResponse().addCookie(cookie);
+                logger.info("refresh token = " + refreshToken);
+
+            }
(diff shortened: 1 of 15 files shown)


src/main/resources/config/application.yml
```diff
@@ -13,6 +13,28 @@
...
+    host:
+        max-total-connections: 1000
+        max-per-route-connections: 100
+    semaphore:
+        max-semaphores: 500
+    retryable: true
+    routes:
+        ManagementPortal:
+            path: /oauthserver/oauth/**
+            url: http://localhost:8080/oauth
```
dennyverbeeck: For `prod` profile, this needs to be `url: http://localhost:8080/managementportal/oauth` due to the context path being set.

src/main/java/org/radarcns/management/filters/OAuth2TokenRequestPostZuulFilter.java
```diff
@@ -0,0 +1,91 @@
...
+                    .readValue(responseBody, new TypeReference<Map<String, Object>>() {
+                    });
+                final String refreshToken = responseMap.get("refresh_token").toString();
+                responseMap.remove("refresh_token");
+                responseBody = mapper.writeValueAsString(responseMap);
+
+                final Cookie cookie = new Cookie(REFRESH_TOKEN_COOKIE, refreshToken);
+                cookie.setHttpOnly(true);
+                cookie.setSecure(true);
+                cookie.setPath(ctx.getRequest().getContextPath() + "oauthserver/oauth/token");
```
dennyverbeeck: We need a leading slash here `/oauthserver/...`

src/main/java/org/radarcns/management/filters/OAuth2TokenRequestPostZuulFilter.java
```diff
@@ -0,0 +1,91 @@
...
+                cookie.setMaxAge(this.managementPortalProperties.getFrontend().getSessionTimeout()); // 30 minites
+                ctx.getResponse().addCookie(cookie);
+                logger.info("refresh token = " + refreshToken);
+
+            }
+            if (requestURI.contains("oauth/token") && requestMethod.equals("DELETE")) {
+                final Cookie cookie = new Cookie(REFRESH_TOKEN_COOKIE, "");
+                cookie.setMaxAge(0);
+                cookie.setPath(ctx.getRequest().getContextPath() + "oauthserver/oauth/token");
```
dennyverbeeck: Leading slash here as well `/oauthserver/...`

src/main/java/org/radarcns/management/filters/OAuth2TokenRequestPostZuulFilter.java
```diff
@@ -0,0 +1,91 @@
...
+                final Map<String, Object> responseMap = mapper
+                    .readValue(responseBody, new TypeReference<Map<String, Object>>() {
+                    });
+                final String refreshToken = responseMap.get("refresh_token").toString();
+                responseMap.remove("refresh_token");
+                responseBody = mapper.writeValueAsString(responseMap);
+
+                final Cookie cookie = new Cookie(REFRESH_TOKEN_COOKIE, refreshToken);
+                cookie.setHttpOnly(true);
+                cookie.setSecure(true);
```
dennyverbeeck: Perhaps we can check the currently active profile (autowire the Spring `Environment` into this class and check it there) and only set the secure flag on the production profile. Since on dev profile we will probably not have HTTPS and therefore the cookie will not be sent from the browser to the server.

src/main/java/org/radarcns/management/filters/OAuth2TokenRequestPreZuulFilter.java
```diff
@@ -0,0 +1,80 @@
...
+
+    static final String REFRESH_TOKEN_COOKIE = "rft";
+    @Autowired
+    private  ManagementPortalProperties managementPortalProperties;
+
+    Logger logger = LoggerFactory.getLogger(OAuth2TokenRequestPreZuulFilter.class);
+    @Override
+    public Object run() {
+        RequestContext ctx = RequestContext.getCurrentContext();
+        if (ctx.getRequest().getRequestURI().contains("/oauth/token")) {
```
dennyverbeeck: You can move this check to `shouldFilter()`, since the method exists for this purpose.

src/main/webapp/app/blocks/interceptor/auth.interceptor.ts
```diff
@@ -2,18 +2,22 @@ import { Observable } from 'rxjs/Observable';
...
+        private injector: Injector,
     ) {
         super();
     }
 
     requestIntercept(options?: RequestOptionsArgs): RequestOptionsArgs {
-        const token = this.localStorage.retrieve('authenticationToken') || this.sessionStorage.retrieve('authenticationToken');
+        // retrieve token from cookie
+        const cookieService = this.injector.get(CookieService);
+        const token : any = cookieService.getObject(AUTH_TOKEN_COOKIE);
```
dennyverbeeck: Perhaps we can take this opportunity to change this bit to use the `auth-oauth2.service`'s `getToken()` method instead.

src/main/java/org/radarcns/management/filters/OAuth2TokenRequestPostZuulFilter.java
nivemaham: Indeed. I tested it. Since it added the cookie on `HTTP` also i left it as it is. I will add that check  anyway.

src/main/webapp/app/blocks/interceptor/auth.interceptor.ts
nivemaham: I will check. The reason i used CookieService is to avoid cyclic dependency, since auth-oauth2.service depends on `Http` and this is an HTTP interceptor. But i think we could still use the injector to get the instance of the service. I will have a look.

PR #130, opened 2017-11-14
Oauth client registration

### LINKED ISSUE #18: Allow dynamic oauth2 client registration with variable grant-types
Currently, we support only one oauth2 client in the system. According to current situation, clients who wants to request tokens should share the client-id and client-secret internally. This is acceptable for trusted clients. This should be extended to allow multiple Oauth2 client registration like other identity provides, with custom client-id, client-secret, and scope. 
This should be discussed in detail.

dennyverbeeck:
- Add API endpoints and UI components for CRUD operations on oauth clients
- Add additional keyword `protected` for the `additional_information` property of OAuth clients. If this key is set to `true`, the OAuth client will not be able to be deleted or modified in any way with by using the UI or API. This is to protect an OAuth client like our MP UI from becoming accidentally misconfigured and as a result being locked out.
- Change the mechanism for loading OAuth clients, no longer a liquibase load. This allows us to supply client secrets in plain text and have them be run through the password encoder before load. The client for the frontend is loaded always and can be configured using the `managementportal.frontend.*` settings. Other clients can be supplied using an identical CSV as was the case with liquibase. Location of this file can be configured using the `managementportal.oauth.clientsFile` property, default location for production profile is `/mp-includes/config/oauth_client_details.csv` and for development profile is `src/main/docker/etc/config/oauth_client_details.csv`. Remember Spring allows these properties to be overridden using environment variables through relaxed binding, e.g. `MANAGEMENTPORTAL_OAUTH_CLIENTS_FILE` will override the setting from the application.yml file.
- **API CHANGE**: `/api/oauthclients` is now at `/api/oauth-clients`

Closes #18


src/main/java/org/radarcns/management/web/rest/OAuthClientsResource.java
```diff
@@ -80,20 +93,110 @@
...
+     * @param clientDetailsDTO The oauth client to be registered
+     * @return a response indicating success or failure
+     * @throws URISyntaxException if there was a problem formatting the URI to the new entity
+     */
+    @PostMapping("/oauthclients")
+    @Timed
+    public ResponseEntity<ClientDetailsDTO> createOAuthClient(@RequestBody ClientDetailsDTO
+            clientDetailsDTO) throws URISyntaxException {
+        checkPermission(getJWT(servletRequest), OAUTHCLIENTS_CREATE);
+        ClientDetails details = clientDetailsMapper
```
nivemaham: Should check whether  `client-id` is null and  whether there a client already exist with current `client-id`.  Similar to 
```
        if (sourceDTO.getId() != null) {
            return ResponseEntity.badRequest().headers(HeaderUtil.createFailureAlert(ENTITY_NAME, "idexists", "A new source cannot already have an ID")).body(null);
        } else if (sourceDTO.getSourceId() != null) {
            return ResponseEntity.badRequest()
                .headers(HeaderUtil.createFailureAlert(ENTITY_NAME, "sourceIdExists", "A new source cannot already have a Source ID"))
                .body(null);
        } else if (sourceRepository.findOneBySourceName(sourceDTO.getSourceName()).isPresent()) {
            return ResponseEntity.badRequest()
                .headers(HeaderUtil.createFailureAlert(ENTITY_NAME, "sourceNameExists", "Source name already in use"))
                .body(null);
        }
```

src/main/java/org/radarcns/management/web/rest/OAuthClientsResource.java
```diff
@@ -80,20 +93,110 @@
...
+    private static final String PROTECTED_KEY = "protected";
+
+    /**
+     * GET /api/oauthclients
+     *
+     * Retrieve a list of currently registered OAuth clients.
+     *
+     * @return the list of registered clients as a list of {@link ClientDetailsDTO}
+     */
     @GetMapping("/oauthclients")
```
nivemaham: I guess `oauth-clients` is the correct convention.

src/main/java/org/radarcns/management/web/rest/OAuthClientsResource.java
```diff
@@ -80,20 +93,110 @@
...
+     * @param id the client id for which to fetch the details
+     * @return the client as a {@link ClientDetailsDTO}
+     */
+    @GetMapping("/oauthclients/{id}")
     @Timed
     public ResponseEntity<ClientDetailsDTO> getOAuthClientById(@PathVariable("id") String id) {
         checkPermission(getJWT(servletRequest), OAUTHCLIENTS_READ);
-        return ResponseEntity.ok().body(
-            clientDetailsMapper.clientDetailsToClientDetailsDTO(getOAuthClient(id)));
+        return ResponseEntity.ok().body(clientDetailsMapper
```
nivemaham: Validate whether id is empty

src/main/webapp/app/entities/oauth-client/oauth-client-dialog.component.html
```diff
@@ -0,0 +1,75 @@
...
+        <div class="alert alert-danger" *ngIf="protectedClient">
+            <strong>This is a protected client.</strong> You will not be able to make any changes to it.
+        </div>
+        <div class="form-group">
+            <label class="form-control-label" for="id" jhiTranslate="managementPortalApp.oauthClient.clientId">Client ID</label>
+            <input *ngIf="!newClient" type="text" class="form-control" id="id" name="id" [(ngModel)]="client.clientId" readonly />
+            <input *ngIf="newClient" type="text" class="form-control" id="id" name="id" [(ngModel)]="client.clientId" required />
+        </div>
+        <div class="form-group">
+            <label class="form-control-label" for="" jhiTranslate="managementPortalApp.oauthClient.clientSecret">Client Secret</label>
```
nivemaham: Not sure whether its possible. Can we change the values into `password` mode when the user switches to other UI element or have a hide button?

src/main/webapp/app/entities/oauth-client/oauth-client-dialog.component.html
```diff
@@ -0,0 +1,75 @@
...
+            <div class="input-group">
+                <input type="text" class="form-control" id="secret" name="clientSecret" [(ngModel)]="client.clientSecret" [disabled]="protectedClient" [required]="newClient" />
+                <span class="input-group-btn"><button class="btn btn-secondary" type="button" (click)="generateRandomSecret()"><span class="fa fa-magic">&nbsp;</span>Randomize</button></span>
+            </div>
+            <small id="scopeHelp" class="form-text text-muted" *ngIf="!newClient">If this field is left empty the secret will not be modified.</small>
+        </div>
+        <div class="form-group">
+            <label class="form-control-label" for="" jhiTranslate="managementPortalApp.oauthClient.scope">Scope</label>
+            <input type="text" class="form-control" id="scope" name="scope" [(ngModel)]="scopeList" [disabled]="protectedClient" required />
```
nivemaham: Again trying to be user friendly. Not sure about the complexity. Can we limit the scopes, or provide the available scopes from where you can choose? writing all of this by hand is going to be difficult.

src/main/webapp/app/entities/oauth-client/oauth-client-dialog.component.html
```diff
@@ -0,0 +1,75 @@
...
+            <label class="form-control-label" for="" jhiTranslate="managementPortalApp.oauthClient.authorizedGrantTypes">Grant Types</label>
+            <div class="form-check" *ngFor="let grant of grantTypes">
+                <label class="form-check-label">
+                    <input type="checkbox" class="form-check-input" name="grantTypes" [checked]="grant.authorized" (change)="grant.authorized = !grant.authorized" [disabled]="protectedClient" /> {{grant.name}}
+                </label>
+            </div>
+        </div>
+        <div class="form-group">
+            <label class="form-control-label" for="" jhiTranslate="managementPortalApp.oauthClient.autoApprove">Auto Approve</label>
+            <input type="text" class="form-control" id="autoApprove" name="autoApprove" [(ngModel)]="autoApproveScopeList" [disabled]="protectedClient" />
```
nivemaham: I see that the server doesn't accept if we try to give a scope that is not provided earlier. However, there is no message or feedback given to user when they try to do that.  I think at least it is a good to let the user choose from the scopes that are already provided and it should be easy to implement since we have two way binding.

src/main/webapp/app/entities/oauth-client/oauth-client-dialog.component.html
```diff
@@ -0,0 +1,75 @@
...
+            <input type="text" class="form-control" id="accessTokenValidity" name="accessTokenValidity" [(ngModel)]="client.accessTokenValidity" [disabled]="protectedClient" required pattern="\d+" />
+        </div>
+        <div class="form-group">
+            <label class="form-control-label" for="" jhiTranslate="managementPortalApp.oauthClient.refreshTokenValidity">Refresh Token Validity</label>
+            <input type="text" class="form-control" id="refreshTokenValidity" name="refreshTokenValidity" [(ngModel)]="client.refreshTokenValidity" [disabled]="protectedClient" required pattern="\d+" />
+        </div>
+        <div class="form-group">
+            <label class="form-control-label" for="" jhiTranslate="managementPortalApp.oauthClient.additionalInformation">Additional Information</label>
+            <div class="form-check">
+                    <label class="form-check-label">
```
nivemaham: I like the idea of additional information. Can we also add `protected-clients` here. so that we can have secured clients too.

src/main/webapp/app/entities/oauth-client/oauth-client-dialog.component.ts
```diff
@@ -0,0 +1,186 @@
...
+    protectedClient: boolean;
+
+    constructor(
+        public activeModal: NgbActiveModal,
+        private jhiLanguageService: JhiLanguageService,
+        private alertService: AlertService,
+        private oauthClientSerivce: OAuthClientService,
+        private eventManager: EventManager
+    ) {
+        this.jhiLanguageService.setLocations(['oauthclient']);
```
nivemaham: should be `this.jhiLanguageService.setLocations(['oauthClient']);`

src/main/java/org/radarcns/management/config/OAuth2ServerConfiguration.java
```diff
@@ -194,7 +197,8 @@ public TokenStore tokenStore() {
         public void configure(AuthorizationServerSecurityConfigurer oauthServer) throws Exception {
             oauthServer.allowFormAuthenticationForClients()
                        .checkTokenAccess("isAuthenticated()")
-                       .tokenKeyAccess("isAnonymous() || isAuthenticated()");
+                       .tokenKeyAccess("isAnonymous() || isAuthenticated()")
+                       .passwordEncoder(new BCryptPasswordEncoder());
```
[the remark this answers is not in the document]
dennyverbeeck: This is a BCrypt hash with a cost factor of 10. 10 is the recommended minimum, we can indeed look into increasing the cost factor. It uses random salts and bcrypt is a non-reversible hash so a standard decoder should not be possible.

src/main/java/org/radarcns/management/web/rest/OAuthClientsResource.java
dennyverbeeck: When `id` is left empty, the request will be routed to `/api/oauth-clients` and return all clients. An unknown client will throw a `CustomNotFoundException` and thus use the `ExceptionTranslator` to build a 404 response.

nivemaham: This it should be `@RequestParam`. I don't think a path parameter can be empty. If it can be empty there is no point of having it as path parameter. There is also another api to do that. Following the practice in all the entities i would say it makes sense to have the validation.

src/main/java/org/radarcns/management/web/rest/OAuthClientsResource.java
blootsvoets: or even just `clients`?

src/main/java/org/radarcns/management/web/rest/errors/ExceptionTranslator.java
```diff
@@ -79,7 +79,7 @@ public ParameterizedErrorVM processParameterizedConflict(CustomConflictException
     }
 
     @ExceptionHandler(AccessDeniedException.class)
-    @ResponseStatus(HttpStatus.FORBIDDEN)
+    @ResponseStatus(HttpStatus.UNAUTHORIZED)
```
blootsvoets: Forbidden seems appropriate. According to the `AccessDeniedException` javadoc:
> Thrown if an Authentication object does not hold a required authority.

So the person is authenticated but forbidden. See also: http://www.dirv.me/blog/2011/07/18/understanding-403-forbidden/index.html, otherwise a `WWW-Authenticate` header should also be added.

src/main/webapp/app/entities/oauth-client/oauth-client-dialog.component.ts
blootsvoets: note that here and in other places: a dialog should generally use `addLocation`, a page could use `setLocation`. This prevents other translations on the page to be lost.

src/main/java/org/radarcns/management/web/rest/OAuthClientsResource.java
nivemaham: Sorry for the confusion. It is not necessary . I got confused with the below validation. For `PUT` we need additional validation before we proceed.

dennyverbeeck: Indeed, will add it!

src/main/java/org/radarcns/management/web/rest/OAuthClientsResource.java
dennyverbeeck: Hmm, i think `clients` is too general, with `oauth-clients` there is no ambiguity as to what we are referring to.

src/main/java/org/radarcns/management/web/rest/errors/ExceptionTranslator.java
dennyverbeeck: Indeed, reverting this change

dennyverbeeck:
Thanks @nivemaham and @blootsvoets for your valuable comments! I'll update the PR message to explicitly mention the `/api/oauthclients` to `/api/oauth-clients` change. I'll be pushing new commits with the changes shortly.

src/main/webapp/app/entities/oauth-client/oauth-client-dialog.component.ts
```diff
@@ -0,0 +1,186 @@
...
+        this.grantTypes = [];
+        this.availableGrants.forEach((grant) => {
+            this.grantTypes.push({"name": grant, "authorized": this.client.authorizedGrantTypes.indexOf(grant) > -1});
+        });
+        // transform array of scopes to a comma seperated string so we can bind a single text box to it
+        this.scopeList = this.client.scope.join(", ");
+        // transform array of resources to a comma seperated string so we can bind a single text box to it
+        this.resourcesList = this.client.resourceIds.join(", ");
+        // transform array of auto-approve scopes to a comma seperated string so we can bind a single text box to it
+        this.autoApproveScopeList = this.client.autoApproveScopes.join(", ");
```
nivemaham: Add ` this.autoApproveScopeList = this.client.autoApproveScopes? this.client.autoApproveScopes.join(", "):"" ;`
The UI crashes when autoApproveScopes is null.

src/main/java/org/radarcns/management/web/rest/OAuthClientsResource.java
```diff
@@ -194,6 +196,15 @@
         checkPermission(getJWT(servletRequest), OAUTHCLIENTS_CREATE);
         // check if we have an ID field supplied
         checkClientFields(clientDetailsDTO);
+        // check if the client id exists
+        try {
+            ClientDetails existing = clientDetailsService.loadClientByClientId(clientDetailsDTO
+                    .getClientId());
```
nivemaham: FYI, #137 is something we should keep in mind.

PR #131, opened 2017-11-17
Refactor liquibase and rename sensordata

dennyverbeeck:
This is a combination of refactoring the initial database schema definition files in liquibase, and @nivemaham 's work on renaming the `SensorData` entity to `SourceData`.


src/main/resources/config/liquibase/changelog/00000000000000_initial_schema.xml
```diff
@@ -88,50 +88,38 @@
...
-        <!--<loadData encoding="UTF-8"-->
-                  <!--file="config/liquibase/users_authorities.csv"-->
-                  <!--separator=";"-->
-                  <!--tableName="radar_user_authority"/>-->
         <createTable tableName="jhi_persistent_audit_event">
             <column name="event_id" type="bigint" autoIncrement="${autoIncrement}">
                 <constraints primaryKey="true" nullable="false"/>
             </column>
-            <column name="principal" type="varchar(50)">
+            <column name="principal" type="varchar(1024)">
```
nivemaham: any reason to have such a big varchar for audit events?

dennyverbeeck: Ah that's a mistake, the `principal` is just the username which is also `varchar(50)`, will revert this line. Thanks!

PR #133, opened 2017-11-17
Improved audits

### LINKED ISSUE #44: Logging operation
We should audit/log all operations that a Management Portal user does. In this way there is a clear understanding of `who did what and when`.

### LINKED ISSUE #106: Check authentication audit events
An `AUTHENTICATION_SUCCESS` audit event is generated every time a protected endpoint is called. When browsing around in the app, this leads to an excessive amount of these events. Preferably, this event should only be logged when a token is obtained (e.g. through a user login, a client_credentials login, or use of a refresh token).

dennyverbeeck:
- Improved display of the audit extra data fields in the UI
- Disabled auditing of every authenticated API request
- Enabled auditing of successful and unsuccessful login attempts
- Enabled auditing of token grants
- Enabled auditing of entity events (create, update and delete)
- Enabled auditing of refresh token generation (Pair app function)
- Enabled auditing of discontinue subject

Closes #106, closes #44


nivemaham:
Haven't started reviewing yet. But, can we add audit events when there is an `AUTHENTICATION_FAILURE`?

dennyverbeeck:
Yep! These are the unsuccessful login attempts i was referring to.

nivemaham:
Sorry. My eyes are skipping lines.

src/main/webapp/app/admin/audits/audit.model.ts
```diff
@@ -1,8 +1,6 @@
-import { AuditData } from './audit-data.model';
-
 export class Audit {
     constructor(
-        public data: AuditData,
+        public data: any,
```
blootsvoets: why not update AuditData?

dennyverbeeck: The audit data was already implemented as a `Map<String, Object>` on the back-end, so I'm not sure if there is a reason to maintain a class for it here. Perhaps instead of `any` a `Map<string, string>` is more appropriate here though.

src/main/java/org/radarcns/management/config/SecurityConfiguration.java
```diff
@@ -34,6 +37,9 @@
 
     private final UserDetailsService userDetailsService;
```
nivemaham: I think a good practice is to either do constructor based DI or using `@AutoWired`. Using solely `@AutoWired` is preferred and it eases testing too.

src/main/java/org/radarcns/management/domain/support/AutowireHelper.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.stereotype.Component;
+
+/**
+ * Created by dverbeec on 17/11/2017.
+ */
+@Component
+public class AutowireHelper implements ApplicationContextAware {
```
nivemaham: Doesn't the ` @Qualifier("nameOfTheInstance")` support this?

nivemaham: I get the reason for custom AutowireHelper. I think it would be good if we also add the limitation of not being able to Autowire beans in JPA Event listeners to the Javadoc.

dennyverbeeck: Good point! Added a brief Javadoc to the class.

src/main/java/org/radarcns/management/config/SecurityConfiguration.java
dennyverbeeck: Indeed, took the liberty change this in the other configuration classes as well

PR #134, opened 2017-11-20
Schema changes

nivemaham:


src/main/java/org/radarcns/management/domain/SourceData.java
```diff
@@ -78,7 +85,7 @@
     @ManyToMany(mappedBy = "sourceData" , fetch = FetchType.EAGER)
     @JsonIgnore
     @Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)
-    private Set<DeviceType> deviceTypes = new HashSet<>();
+    private Set<SourceType> sourceTypes = new HashSet<>();
```
dennyverbeeck: Should this also be many-to-one? Or can a `SourceData` belong to multiple `SourceType`s?

nivemaham: No. Should be ManyToOne. Forgot to change. Will do.

dennyverbeeck:
For reference:

**API CHANGES**
- `GET /api/projects/{projectName}/device-types` -> `GET /api/projects/{projectName}/source-types`
- `/api/sensor-data` -> `/api/source-data`
- `/api/device-types` -> `/api/source-types`
- All field names `deviceType*` renamed to `sourceType*` (this includes bodies of POSTs like `POST /api/subjects/{login}/sources`)
- All field names `sensorData*` renamed to `sourceData*`

Issue #132, opened 2017-11-17
Update database entities for source-type import


nivemaham:
Database schema should be updated according to keep the information from radar-schemas

Issue #125, opened 2017-11-09
Device type creation: Dynamic source registration


dennyverbeeck:
Can not set the dynamic source registration when creating a new device type, either when creating or editing. There should be a checkbox to enable or disable dynamic source registration.

dennyverbeeck:
Correction: clicking the label 'Dynamic source registration' and then saving actually toggles the state, so there might be display issue.

nivemaham:
@dennyverbeeck i tested this on latest dev and i do not see any problems. Clicking on the Dynamic source registration checkbox should change the state, so should be the display.

PR #153, opened 2017-11-27
Source type import rebase

nivemaham:
Implemented a source-type importer from catalog server.
The `serverUrl` and `enableAutoImport` properties need to be configured according to the deployment setup in `application-*.yml`.
If the source-type import is enabled, the service will try to load all the source-types available in catalog-server from Radar-schemas.


src/main/java/org/radarcns/management/ManagementPortalTestApp.java
```diff
@@ -0,0 +1,88 @@
...
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.FilterType;
+import org.springframework.core.env.Environment;
+
+@ComponentScan(excludeFilters = @ComponentScan.Filter(type = FilterType.ASSIGNABLE_TYPE, value = SourceTypeLoader.class))
+@EnableZuulProxy
+@EnableAutoConfiguration(exclude = {MetricFilterAutoConfiguration.class, MetricRepositoryAutoConfiguration.class})
+@EnableConfigurationProperties({LiquibaseProperties.class, ApplicationProperties.class ,
+    ManagementPortalProperties.class})
+public class ManagementPortalTestApp {
```
dennyverbeeck: Perhaps it would be good to add a Javadoc comment to explain how this class is different from the regular `ManagementPortalApp` and why it is necessary.

src/main/java/org/radarcns/management/web/rest/util/HttpUtil.java
```diff
@@ -0,0 +1,23 @@
...
+public class HttpUtil {
+
+    public static boolean isReachable(URL urlServer) {
+        try {
+            HttpURLConnection urlConn = (HttpURLConnection) urlServer.openConnection();
+            urlConn.setConnectTimeout(3000); //<- 3Seconds Timeout
+            urlConn.connect();
+            return urlConn.getResponseCode() == 200;
+        } catch (IOException e) {
+            e.printStackTrace();
```
dennyverbeeck: Perhaps it is sufficient to do a `log.warn("Server {} is unreachable: {}", urlServer.toString(), e.getMessage())`, a stacktrace less nice to deal with in the logs.

nivemaham:
Fixes #132 , Fixes #125

PR #155, opened 2017-11-27
Uriencode frontend requests

### LINKED ISSUE #135: URLEncode request parameters on frontend
Path variables for requests to the API should be properly URL encoded. E.g. a project called `radar/1`, will have the slash URL encoded in the routerLinks (i.e. `radar%2F1`), however the request to the API looks like `GET /api/project/radar/1` which of course yields a 404.

dennyverbeeck:
All entity fields that are part of the entity's identifier now need to adhere to the regex `^[_'.@A-Za-z0-9-]*$`.

- Frontend correctly encodes special characters in path variables
- Add form validation and messages to entity identifier fields
- Add validation on the relevant domain class fields
- Add exception handler for `TransactionSystemException`
- `Constants` class moved to `radar-auth`

Closes #135

### DIFF
--- a/src/main/java/org/radarcns/management/domain/Source.java
+++ b/src/main/java/org/radarcns/management/domain/Source.java
@@ -3,6 +3,7 @@ package org.radarcns.management.domain;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import org.hibernate.annotations.Cache;
 import org.hibernate.annotations.CacheConcurrencyStrategy;
+import org.radarcns.auth.config.Constants;
 
 import javax.persistence.CollectionTable;
 import javax.persistence.Column;
@@ -20,6 +21,7 @@ import javax.persistence.PrePersist;
 import javax.persistence.SequenceGenerator;
 import javax.persistence.Table;
 import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
 import java.io.Serializable;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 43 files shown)


src/main/java/org/radarcns/management/domain/Source.java
```diff
@@ -48,9 +50,11 @@
     private UUID sourceId;
 
     @NotNull
+    @Pattern(regexp = Constants.ENTITY_ID_REGEX)
     @Column(name = "source_name", nullable = false, unique = true)
     private String sourceName;
 
+    @Pattern(regexp = Constants.ENTITY_ID_REGEX)
```
nivemaham: Do we need this?

src/main/java/org/radarcns/management/domain/SourceType.java
```diff
@@ -55,10 +60,12 @@
     private String appProvider;
 
     @NotNull
+    @Pattern(regexp = Constants.ENTITY_ID_REGEX)
     @Column(name = "model", nullable = false )
     private String model;
 
     @NotNull
+    @Pattern(regexp = Constants.ENTITY_ID_REGEX)
```
nivemaham: Should we add these `@pattern ` annotations to other resources as well, since we have a pattern requirement now.

src/main/java/org/radarcns/management/domain/Source.java
dennyverbeeck: Indeed, probably not, removing this.

src/main/java/org/radarcns/management/domain/SourceType.java
dennyverbeeck: I added them to all fields that are part of the entity's identifier. E.g. `sourceName` for a `Source`, `projectName` for a `Project`, or `producer`, `model` and `catalogVersion` in the case of `SourceType`. Do you have other suggestions, or did i miss any?

nivemaham: SourdeDataName for sourceData

nivemaham: My bad. SourceDataName is there

PR #157, opened 2017-11-28
Security update

### LINKED ISSUE #144: Review security implementation
Authorization should be based on a token having the correct scope AND the user having the correct role. Currently it's one or the other.

dennyverbeeck:
JWT's now need the correct scope in addition to the correct user authority. This allows a user to authorize certain applications for certain scopes, instead of the application automatically 'inheriting' all user authorities.

Closes #144

### DIFF
--- a/radar-auth/src/main/java/org/radarcns/auth/authorization/RadarAuthorization.java
+++ b/radar-auth/src/main/java/org/radarcns/auth/authorization/RadarAuthorization.java
@@ -22,6 +22,9 @@ public class RadarAuthorization {
     public static final String AUTHORITIES_CLAIM = "authorities";
     public static final String ROLES_CLAIM = "roles";
     public static final String SCOPE_CLAIM = "scope";
+    public static final String SOURCES_CLAIM = "sources";
+    public static final String GRANT_TYPE_CLAIM = "grant_type";
+    private static final String CLIENT_CREDENTIALS = "client_credentials";
 
     /**
      * Check if the user authenticated with the given token has the given permission. Not taking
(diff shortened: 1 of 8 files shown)


radar-auth/src/main/java/org/radarcns/auth/authorization/RadarAuthorization.java
```diff
@@ -80,35 +69,25 @@ public static void checkPermissionOnSubject(DecodedJWT token, Permission permiss
...
-                    projectName, subjectName));
-        } else {
-            // otherwise we have other roles and we should check on a project level
-            checkPermissionOnProject(token, permission, projectName);
-        }
-    }
+        checkScope(token, permission);
+        if (!isClientCredentials(token)) {
+            // we're allowed to read our own data
+            if (token.getSubject().equals(subjectName) && Permissions.allowedAuthorities(permission)
```
nivemaham: Is this api used by our clients? Since `SubjectDTO` exposed quite some data about the sources, projects, external ids etc. I wonder, what would if by any chance a valid refresh-token is exposed, which i think could be possible then they will get all the above details. I would say subject shouldn't be able to read on the other hand i also think, `subject/{login/sources` is used for `ROLE_PARTICIPANT` for `pRMT`. Maybe we can think of a processed meta-data for the future.

dennyverbeeck: For future reference: after offline discussion we decided to keep the current permissions for `ROLE_PARTICIPANT`, as not to break the apps. We'll open a discussion with the wider group to get a consensus on the permissions a participant should and should not have at a later time.

PR #161, opened 2017-11-28
Add production e2e tests

dennyverbeeck:
Adds extra steps in travis to 
- also do e2e tests on a production backend, since the settings are different there
- build the Dockerfile in the project root when on a branch with the name `release-*`, so we can find issues with the docker build before we tag a release


src/test/bash/run-prod-e2e.sh
```diff
@@ -0,0 +1,32 @@
...
+#
+# Then we start up a docker stack with a postgres server since production mode is configured for a
+# postgres database instead of in-memory database.
+
+set -ev
+
+sed -i "s|new plugin.BaseHrefWebpackPlugin({ baseHref: '/' })|new plugin.BaseHrefWebpackPlugin({ baseHref: '/managementportal/' })|" webpack/webpack.dev.js
+sed -i "s|baseUrl: 'http://localhost:8080/',|baseUrl: 'http://localhost:8080/managementportal/',|" src/test/javascript/protractor.conf.js
+sed -i "s|contexts: prod|contexts: dev|" src/main/resources/config/application-prod.yml # set liquibase context to dev so it loads demo data
+./gradlew bootRepackage -Pprod buildDocker -x test
```
nivemaham: -Pprod will load `prod` profile. then why do we change the `webpack.dev.js`?

PR #168, opened 2017-12-06
Update docs

### LINKED ISSUE #156: Create swagger documentation on release
Swagger documentation can for example be put in a doc folder or a separate zip file in GitHub releases, or part of GitHub pages. That way it is easy to look up the valid API calls for a given release.

dennyverbeeck:
- A `gh-pages` branch was added from which our Github pages site is built.
- Update the Readme to  reference the documentation on Githb pages
- Add a script to build javadoc and API documentation based on the swagger.json file and ReDoc
- Documentation based on v0.3.1 is already available [here](https://radar-cns.github.io/ManagementPortal/)

Closes #156


README.md
```diff
@@ -17,22 +36,58 @@ keytool -genkey -alias selfsigned -keyalg RSA -keystore src/main/docker/etc/conf
 **Make sure the key password and store password are the same!** This is a requirement for Spring Security.
 
 2. Then, make sure [Docker][] and [Docker-Compose][] are installed on your system.
-3. Finally, we can start the stack with `docker-compose -f src/main/docker/app.yml up -d`.
+3. Finally, we can start the stack with `docker-compose -f src/main/docker/management-portal.yml up -d`.
```
nivemaham: Docker setup is optional for MP. It should be covered under docker.

README.md
```diff
@@ -17,22 +36,58 @@ keytool -genkey -alias selfsigned -keyalg RSA -keystore src/main/docker/etc/conf
...
+The table below lists the variables that are most likely in need of change when deploying Management
+Portal. You can find the complete configuration
+in the [application.yml](src/main/resources/config/application.yml) and 
+[application-prod.yml](src/main/resources/config/application-prod.yml) files. See
+[Spring external configuration](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html)
+for other options on overriding the default configuration.
+
+| Variable                                                   | Default value                                       | Description                                                                                            |
+|------------------------------------------------------------|-----------------------------------------------------|--------------------------------------------------------------------------------------------------------|
+| `SPRING_DATASOURCE_URL`                                    | `jdbc:postgresql://localhost:5432/managementportal` | URL for the database to be used                                                                        |
```
nivemaham: This should be moved under docker section. Configuring environment.properties is adequate for standard config

build.gradle
```diff
@@ -30,7 +30,7 @@ plugins {
 
 allprojects {
     group 'org.radarcns'
-    version '0.3.2-SNAPSHOT'
+    version '0.3.1'
```
nivemaham: why are we falling back with the earlier version?

README.md
nivemaham: The Dockerfile under docker folder doesn't do the same as the one under the root directory. It in fact doesn't work as expected. If we have to fully rely on test automation and CI we have, I would recommend having a single dockerfile and make sure it works as expected in all cases.

dennyverbeeck: The Dockerfile in the docker folder is meant for the gradle `buildDocker` task and its purpose is to build an image based on the war generated with `bootRepackage`, i.e. a quick build from an existing compiled file.

The Dockerfile in the project root is meant for Dockerhub automated builds. Since the pre-built war is not available in that scenario this Dockerfile includes the compilation step.

I agree the Readme in general could benefit from some restructuring. Should we modify the quickstart section to explain using `gradlew bootRun`, i.e. `dev` profile with demo data; and have a separate section for all the Docker related information, i.e. `prod` profile and associated configuration?

build.gradle
dennyverbeeck: Not sure how this happened, it's a mistake indeed, good catch!

PR #170, opened 2017-12-11
Improvements

### LINKED ISSUE #143: Improve display for OAuth client with large number of scopes
An OAuth client with large number of scopes will take up a lot of screen space. A UI feature for collapsing and expanding such clients is preferable.

nivemaham:
Added an expand -> collapse feature when there is large number of scopes for a client
Added conversion of seconds to other units of time 

Fixes #143

### DIFF
--- a/src/main/webapp/app/shared/show-more/show-more.component.ts
+++ b/src/main/webapp/app/shared/show-more/show-more.component.ts
@@ -1,0 +1,56 @@
+import {Component, Input, OnInit} from '@angular/core';
+
+@Component({
+    selector: 'show-more',
+    template: `
+        <div >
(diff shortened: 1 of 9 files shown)


src/main/webapp/app/shared/show-more/show-more.component.ts
```diff
@@ -0,0 +1,56 @@
...
+        div.collapsed {
+            overflow: hidden;
+        }
+    `]
+})
+/**
+ * This component collapses if number of array items are more than 10 and allows to expand
+ * and collapse in the view.
+ */
+export class ShowMoreComponent implements OnInit{
```
dennyverbeeck: This is a nice component, we might want to reuse it in other places as well. Instead of talking about `scopes` we could talk about `elements` to make the naming more general; and make the span class that is now `badge badge-primary` an input parameter of the component as well.

PR #171, opened 2017-12-13
Add style definitions to main project

dennyverbeeck:
This allows Codacy to check the main project code base for style as well. The `ignoreFailures` property has been set to `true` so tests won't fail because of incorrect style. We might enable it later when we have synced the complete codebase to the style.

### DIFF
--- a/src/main/java/org/radarcns/management/web/rest/OAuthClientsResource.java
+++ b/src/main/java/org/radarcns/management/web/rest/OAuthClientsResource.java
@@ -206,8 +206,10 @@ public class OAuthClientsResource {
         try {
             ClientDetails existing = clientDetailsService.loadClientByClientId(clientDetailsDTO
                     .getClientId());
-            throw new CustomConflictException("An OAuth client with that ID already exists",
+            if(Objects.nonNull(existing)) {
+                throw new CustomConflictException("An OAuth client with that ID already exists",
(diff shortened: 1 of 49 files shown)


dennyverbeeck:
There are currently, on this branch, 1113 style violation is 139 files :stuck_out_tongue: so i guess we best start working on it on package by package basis and do a style PR by package or something along those lines.

src/main/java/org/radarcns/management/web/rest/OAuthClientsResource.java
```diff
@@ -206,8 +206,10 @@
         try {
             ClientDetails existing = clientDetailsService.loadClientByClientId(clientDetailsDTO
                     .getClientId());
-            throw new CustomConflictException("An OAuth client with that ID already exists",
+            if(Objects.nonNull(existing)) {
```
dennyverbeeck: We don't necessarily need that if statement here. If the client does not exist the `NoSuchClientException` will be thrown and this line won't be reached. If the client exists the exception won't be thrown and this line will be reached, and thus throwing the `CustomConflictException`.

nivemaham: Indeed. But having an unused local variable is a violation. Stupid way of fixing that violation. Pushed a fix now it.

PR #175, opened 2017-12-15
Notifications

### LINKED ISSUE #137: Throwing exceptions doesn't show the alert message on the front end.
We need to find a unique way of addressing exceptions from back-end to clients. Throwing custom exceptions give well formatted HTTP response, however it does not show the alert message in front-end app. We are using bad-request with HTTP headers and custom-exceptions in various places. We should find one solution that would work for all and stick to that in all cases.

dennyverbeeck:
Fixes related to the notification system and translations:

- Create, update and delete notifications now correctly show up in the UI
- Improve notifications and errors throughout the application
- Improve notification display so that it no longer overlaps with other elements on the screen
- Add test to check for consistent translation files
- Fix missing translations in `nl`
- Add some more e2e tests for create, update and delete operations, also checking for the notifications
- Fix an issue that causes the show-more component to expand when an entity table gets updated

Closes #137


src/main/java/org/radarcns/management/web/rest/OAuthClientsResource.java
```diff
@@ -352,7 +331,7 @@ private void checkProtected(ClientDetails details) {
         Map<String, Object> info = details.getAdditionalInformation();
         if (Objects.nonNull(info) && info.containsKey(PROTECTED_KEY)
                 && info.get(PROTECTED_KEY).toString().equalsIgnoreCase("true")) {
-            throw new CustomParameterizedException("Modification of a protected OAuth client is "
+            throw new AccessDeniedException("Modification of a protected OAuth client is "
```
nivemaham: I think `InvalidAccessException` is more suitable than `AccessDeniedException` for this case.

src/main/java/org/radarcns/management/web/rest/RoleResource.java
```diff
@@ -135,4 +139,13 @@
...
     }
 
+    /**
+     * Create a user-friendly display name for a given role. Useful for passing to the
+     * notification system
+     * @param role The role to create a user-friendly display for
+     * @return the display name
+     */
+    private String displayName(RoleDTO role) {
+        return role.getProjectName() + " - " + role.getAuthorityName();
```
nivemaham: Can have `:` instead of `-`. So we ca reuse this method when necessary since it matches existing logics.

src/main/java/org/radarcns/management/web/rest/OAuthClientsResource.java
dennyverbeeck: I'm not seeing an `InvalidAccessException` class in our project or in Spring, is it from another library? You're right, `AccessDenied` might create the expectation that there is a specific role that would allow this operation, which is not the case. So perhaps i'll just revert this change. This would then again yield a 400 Bad request, indicating that there is a problem with the request, not with the permissions. What do you think?

src/main/java/org/radarcns/management/web/rest/RoleResource.java
dennyverbeeck: Good point, will change.

src/main/java/org/radarcns/management/web/rest/OAuthClientsResource.java
nivemaham: Yes. `400` makes more sense. AccessDenied is more suitable for 401 i would say. Yes. I think it is not available in standard packages. But makes more sense. We could create the `InvalidAccessException` or `InvalidActivityException`. So i guess custom exception if necessary and status code `400` is good to go.

PR #179, opened 2017-12-20
Remove unused authorities field from User and drop associated table

### LINKED ISSUE #173: Remove `authorities` from `User` and use `Roles` when relevant.
Currently we have an unused entity field in `User` and a getter that loads data from a different field. This should be handled by the Mapper and there are queries that uses the above method which should use Roles instead

dennyverbeeck:
We still need the `authorities` field in `UserDTO`. The frontend does a request to `/api/account` to get user information, and based on that it decides if you are an admin or not. We might want to change this behavior so that it uses the information in the JWT directly instead of making the extra call.

Closes #173


dennyverbeeck:
Don't want to merge this to master obviously...

nivemaham:
I got a heart attack when i saw 100+ files changes. Of course worried about my merge-conflicts to come :)

PR #181, opened 2017-12-20
Pagination 

### LINKED ISSUE #63: Pageable queries for entities
Newly added entities such as subject, sources, device types should be supported to have pageable queries like users

nivemaham:
Enabled pagination in entity views and for subject and sources in project view
Fixes #63


src/main/java/org/radarcns/management/repository/ProjectRepository.java
```diff
@@ -1,35 +1,42 @@
...
     Project findOneWithEagerRelationshipsByName(@Param("name") String name);
 
     @Query("select project.sourceTypes from Project project WHERE project.id = :id")
     List<SourceType> findSourceTypesByProjectId(@Param("id") Long id);
 
-    @Query("select distinct sourceType from Project project left join project.sourceTypes sourceType "
-        + "where project.id =:id "
-        + "and sourceType.id = :sourceTypeId ")
+    @Query(
+            "select distinct sourceType from Project project "
```
dennyverbeeck: Could probably be on the previous line: `@Query("select distinct ...` just as the other multi-line queries

nivemaham:
@dennyverbeeck Indeed. It is something that i forgot to add. Took a while to fix codacy violations :D

PR #187, opened 2018-01-02
Oauth2 client synchronized

### LINKED ISSUE #185: OAuth2Client http client
Currently `org.radarcns.oauth.OAuth2Client` takes a static singleton HTTP client. The access to this client is not synchronised. It would be preferable to give each instance its own HTTP client. The caller can then share a connection pool, etc. If needed you can still share values like the connection pool or authenticator, if that sharing happens in a synchronised way or is easy to synchronise for external clients. For example, a `public static synchronized` method that creates singleton http client for easy use in the OAuth2Client or as a default.

nivemaham:
Creates separate instances of HTTP-Clients per OAuth2Clients. Correctly updates `Authorization` header.
Fixes #185

### DIFF
--- a/oauth-client-util/src/main/java/org/radarcns/oauth/OAuth2Client.java
+++ b/oauth-client-util/src/main/java/org/radarcns/oauth/OAuth2Client.java
@@ -86,7 +86,7 @@ public class OAuth2Client {
         return currentToken;
     }
 
-    public synchronized OkHttpClient getHttpClient() {
+    public OkHttpClient getHttpClient() {
         if (httpClient == null) {
             // create a client which will supply OAuth client id and secret as HTTP basic authentication
             httpClient = new OkHttpClient.Builder()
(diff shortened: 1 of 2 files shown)


oauth-client-util/src/main/java/org/radarcns/oauth/OAuth2Client.java
```diff
@@ -89,66 +86,44 @@ public OAuth2AccessTokenDetails getAccessToken() throws TokenException {
         return currentToken;
     }
 
-    public OkHttpClient getHttpClient() {
+    public synchronized OkHttpClient getHttpClient() {
```
blootsvoets: if `httpClient` is an instance variable, removing synchronised here would be fine.

oauth-client-util/src/main/java/org/radarcns/oauth/OAuth2Client.java
```diff
@@ -36,7 +33,7 @@
     private Set<String> scope;
     private OAuth2AccessTokenDetails currentToken;
 
-    private static OkHttpClient HTTP_CLIENT;
+    private OkHttpClient HTTP_CLIENT;
```
blootsvoets: perhaps rename to `httpClient`?

nivemaham: Done

PR #191, opened 2018-01-09
Change TokenException to inherit GeneralSecurityException

nivemaham:
Change TokenException to inherit GeneralSecurityException and validae response before processing token response body

### DIFF
--- a/oauth-client-util/src/main/java/org/radarcns/oauth/OAuth2Client.java
+++ b/oauth-client-util/src/main/java/org/radarcns/oauth/OAuth2Client.java
@@ -119,8 +119,16 @@ public class OAuth2Client {
             .build();
 
         // make the client execute the POST request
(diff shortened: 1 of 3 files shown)


oauth-client-util/src/main/java/org/radarcns/oauth/OAuth2Client.java
```diff
@@ -119,8 +119,16 @@ private void getNewToken() throws TokenException {
             .build();
 
         // make the client execute the POST request
-        try (Response response = getHttpClient().newCall(request).execute()) {
-            currentToken = OAuth2AccessTokenDetails.getObject(response);
+        try {
+            Response response = getHttpClient().newCall(request).execute();
```
dennyverbeeck: `Response` implements `AutoCloseable` so we can keep using it in a try-with-resources statement.

Issue #195, opened 2018-01-09
Query custom claims from JWT


blootsvoets:
It would be useful to query MP-custom JWT claims using radar-auth, such as:
- sources
- permissions
- roles
- projects

Perhaps something like
```java
DecodedJWT token = ...;
RadarJwt jwt = new RadarJwt(token)
class Role {
  String project;
  String role;
}
List<String> projects = jwt.getProjects();
List<Role> roles = jwt.getRoles();
Set<Permission> permissions = jwt.getPermissions();
List<String> sources = jwt.getSources();
```

Issue #194, opened 2018-01-09
Add non-throwing authorization


blootsvoets:
Sometimes, an application wants to decide what to do based on whether a user has permission. In this branching logic, it would be nice to have a `boolean RadarAuthorization.hasPermissionOnProject` and equivalents instead of `void RadarAuthorization.checkPermissionOnProject`.

PR #197, opened 2018-01-10
Change NotAuthorizedException's base class to GeneralSecurityException

### LINKED ISSUE #190: Use explicit security exceptions
Radar-auth authentication throw exceptions that extend from `RuntimeException`. This means there is no warning when they are not explicitly caught, and they can't be caught as one general exception. I'd propose to change the base class to [`GeneralSecurityException`](https://docs.oracle.com/javase/7/docs/api/java/security/GeneralSecurityException.html). Then they need to be caught explicitly, and they have a superclass that can be caught. @nivemaham @dennyverbeeck any thoughts?

dennyverbeeck:
Together with PR #191 this closes #190

### DIFF
--- a/radar-auth/src/main/java/org/radarcns/auth/authorization/RadarAuthorization.java
+++ b/radar-auth/src/main/java/org/radarcns/auth/authorization/RadarAuthorization.java
@@ -26,11 +26,14 @@ public class RadarAuthorization {
 
     /**
      * Check if the user authenticated with the given token has the given permission. Not taking
-     * into account project affiliations.
+     * into account project affiliations. Throws a {@link NotAuthorizedException} if the supplied
+     * token does not have the permission.
      * @param token The token of the logged in user
      * @param permission The permission to check
+     * @throws NotAuthorizedException if the supplied token does not have the permission
      */
-    public static void checkPermission(DecodedJWT token, Permission permission) {
+    public static void checkPermission(DecodedJWT token, Permission permission)
+            throws NotAuthorizedException {
         log.debug("Checking permission {} for user {}", permission.toString(), token.getSubject());
         checkScope(token, permission);
         if (!isClientCredentials(token)) {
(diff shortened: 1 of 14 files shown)


radar-auth/src/main/java/org/radarcns/auth/authorization/RadarAuthorization.java
```diff
@@ -26,11 +26,14 @@
 
     /**
      * Check if the user authenticated with the given token has the given permission. Not taking
-     * into account project affiliations.
+     * into account project affiliations. Throws a {@link NotAuthorizedException} if the supplied
```
blootsvoets: This last sentence would not be necessary, it is documented with the `@throws` tag.

src/main/java/org/radarcns/management/service/SubjectService.java
```diff
@@ -126,7 +126,7 @@ private Role getProjectParticipantRole(ProjectDTO projectDTO) {
 
 
     @Transactional
-    public SubjectDTO updateSubject(SubjectDTO subjectDTO) throws IllegalAccessException {
+    public SubjectDTO updateSubject(SubjectDTO subjectDTO) {
```
nivemaham: any idea why we had this exception earlier?

dennyverbeeck:
The `ExceptionTranslator` class already had a handler for `NotAuthorizedException`, so it will still return a `FORBIDDEN` status code. Since the base class for `NotAuthorizedException` changed from unchecked type (`RuntimeException`) to a checked exception, we now need to explicitly add the throws declaration to the controller methods. But it does not affect the `ExceptionTranslator`.

src/main/java/org/radarcns/management/service/SubjectService.java
dennyverbeeck: Going back in the history, it was added in [this commit](https://github.com/RADAR-CNS/ManagementPortal/commit/df0608278860f9f0b27ae94cdae3386553cdf87a#diff-359c75b052347bcf9e55b4901495674eR109), but no other changes were made to the method in that commit. So I assume it was added when working on the method, and that the throwing code got removed again before the commit. In any case i don't see anything there now that would throw an `IllegalAccessException`.

dennyverbeeck:
Indeed, the javadoc will be updated in a later PR i'm working on, related to #194 and #195, so i will go ahead with the merge for this PR now. Thanks!

PR #198, opened 2018-01-10
Refactor radar-auth

### LINKED ISSUE #194: Add non-throwing authorization

### LINKED ISSUE #195: Query custom claims from JWT

dennyverbeeck:
- The token implementation is now wrapped with a `RadarToken` interface
- The new `RadarToken` has methods to check for permissions and query custom claims
- `RadarAuthorization` is still there, but only keeps the `check...` methods
- The public static claim names are moved from `RadarAuthorization` to `JwtRadarToken`
- `Permission` class now has a method to check if a given authority has that permission
- `TokenValidator.validateAccessToken(String)` now returns a `RadarToken` so clients of this library should only have to deal with our wrapper of the token.

Closes #194, closes #195

### DIFF
--- a/radar-auth/src/main/java/org/radarcns/auth/token/AbstractRadarToken.java
+++ b/radar-auth/src/main/java/org/radarcns/auth/token/AbstractRadarToken.java
@@ -1,0 +1,112 @@
+package org.radarcns.auth.token;
+
+import org.radarcns.auth.authorization.AuthoritiesConstants;
+import org.radarcns.auth.authorization.Permission;
+
+import java.util.Collections;
+
+/**
+ * Partial implementation of {@link RadarToken}, providing a default implementation for the three
+ * permission checks.
+ */
+public abstract class AbstractRadarToken implements RadarToken {
+
+    protected static final String CLIENT_CREDENTIALS = "client_credentials";
+
+    @Override
+    public boolean hasPermission(Permission permission) {
+        return hasScope(permission.scopeName())
+                && (isClientCredentials() || hasAuthorityForPermission(permission));
+
+    }
+
+    @Override
+    public boolean hasPermissionOnProject(Permission permission, String projectName) {
+        return hasScope(permission.scopeName())
+                && (isClientCredentials() || hasAuthorityForPermission(permission, projectName));
+    }
+
+    @Override
+    public boolean hasPermissionOnSubject(Permission permission, String projectName,
+            String subjectName) {
+        return hasScope(permission.scopeName())
+                && (isClientCredentials() || hasAuthorityForPermission(permission, projectName,
+                        subjectName));
+    }
+
+    protected boolean hasScope(String scope) {
+        return getScopes().contains(scope);
+    }
+
+    protected boolean isClientCredentials() {
+        return CLIENT_CREDENTIALS.equals(getGrantType());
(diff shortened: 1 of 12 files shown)


src/main/java/org/radarcns/management/security/ClaimsTokenEnhancer.java
```diff
@@ -62,7 +63,7 @@ public OAuth2AccessToken enhance(OAuth2AccessToken accessToken,
                         .map(role -> role.getProject().getProjectName() + ":"
                                 + role.getAuthority().getName())
                         .collect(Collectors.toList());
-                additionalInfo.put(RadarAuthorization.ROLES_CLAIM, roles);
+                additionalInfo.put(JwtRadarToken.ROLES_CLAIM, roles);
```
blootsvoets: I'm wondering if this `additionalInfo` could be something that a `JwtRadarToken` generates. Perhaps with a builder?

radar-auth/src/main/java/org/radarcns/auth/token/AbstractRadarToken.java
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * Check authorities in this token linked to the given project, or not linked to any project
+     * (such as {@code SYS_ADMIN}), for the given permission on the given subject.
+     * @param permission the permission
+     * @param projectName the project name
+     * @param subjectName the subject name
+     * @return {@code true} if any authority contains the permission, {@code false} otherwise
+     */
+    protected boolean hasAuthorityForPermission(Permission permission, String projectName,
```
nivemaham: Looks good to me. Minor comment. Would it help to have different method names than method overloading for these methods. e.g. `hasAuthorityForPermissionInProject` ?

src/main/java/org/radarcns/management/security/ClaimsTokenEnhancer.java
dennyverbeeck: Interesting idea, i think it would make sense indeed to centralize the logic for creating this map in `JwtRadarToken`.

radar-auth/src/main/java/org/radarcns/auth/token/AbstractRadarToken.java
dennyverbeeck: Indeed had to choose between long method names and overloading :stuck_out_tongue: probably a longer method name is still better.

dennyverbeeck:
Thanks for the comments! I'll include them in a seperate PR after all currently open PR's are closed.

PR #199, opened 2018-01-11
Add permission check methods java8

blootsvoets:
Some small Java 8 / notation tweaks:
- helper functions for empty string/list defaults
- directly use unspecified claims: it will return null after conversion to `asList` or `asString`.
- single stream function to parse the roles
- some lambda -> Function::ref changes

### DIFF
--- a/radar-auth/src/main/java/org/radarcns/auth/token/AbstractRadarToken.java
+++ b/radar-auth/src/main/java/org/radarcns/auth/token/AbstractRadarToken.java
@@ -1,5 +1,6 @@
 package org.radarcns.auth.token;
 
+import java.util.Collection;
 import org.radarcns.auth.authorization.AuthoritiesConstants;
 import org.radarcns.auth.authorization.Permission;
 
@@ -50,7 +51,7 @@ public abstract class AbstractRadarToken implements RadarToken {
      */
     protected boolean hasAuthorityForPermission(Permission permission) {
         return getRoles().values().stream()
-                .flatMap(s -> s.stream())
+                .flatMap(Collection::stream)
                 .anyMatch(permission::isAuthorityAllowed)
                 || hasNonProjectRelatedAuthorityForPermission(permission);
     }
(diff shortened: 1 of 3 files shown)


radar-auth/src/main/java/org/radarcns/auth/token/JwtRadarToken.java
```diff
@@ -117,45 +115,26 @@ public String getType() {
     }
 
     private Map<String, List<String>> parseRoles(DecodedJWT jwt) {
-        if (!jwt.getClaims().containsKey(ROLES_CLAIM)) {
-            return Collections.emptyMap();
-        }
-
-        Map<String, List<String>> result = new HashMap<>();
-        List<String[]> parsedRoles = jwt.getClaim(ROLES_CLAIM).asList(String.class).stream()
+        return emptyIfNull(jwt.getClaim(ROLES_CLAIM).asList(String.class)).stream()
```
nivemaham: Would it return an empty map? I only see methods to return empty string or list.

blootsvoets: It would process the empty list and finally return an empty map after the collect method.

radar-auth/src/main/java/org/radarcns/auth/token/AbstractRadarToken.java
```diff
@@ -63,7 +65,7 @@ protected boolean hasAuthorityForPermission(Permission permission) {
      * @return {@code true} if any authority contains the permission, {@code false} otherwise
      */
     protected boolean hasAuthorityForPermission(Permission permission, String projectName) {
-        return getRoles().containsKey(projectName) && getRoles().get(projectName).stream()
+        return getRoles().getOrDefault(projectName, Collections.emptyList()).stream()
```
dennyverbeeck: I can't believe i didn't know about `getOrDefault()`!  Nice :smile:

dennyverbeeck:
I'm sorry @blootsvoets i didn't realize this was merging into the `add-permission-check-methods` branch and i deleted it after it's PR was merged. Thanks for fixing.

PR #201, opened 2018-01-12
Fix style

### LINKED ISSUE #192: Fix code style
We're close to 100% on our code style. Let's fix up everything and disable checkstyle's `ignoreFailures` again.

dennyverbeeck:
- Fix code style in all project files
- Modify checkstyle rules for lowerCamelCase
- Add PMD rule for suggesting class fields that could be final
- Checkstyle will now trigger build failure on style errors

Closes #192


src/main/java/org/radarcns/management/service/RoleService.java
```diff
@@ -64,18 +68,23 @@ public RoleDTO save(RoleDTO roleDTO) {
...
             log.debug("Request to get all Roles");
             result = roleRepository.findAll().stream()
                     .map(roleMapper::roleToRoleDTO)
                     .collect(Collectors.toCollection(LinkedList::new));
         } else if (currentUserAuthorities.contains(AuthoritiesConstants.PROJECT_ADMIN)) {
             log.debug("Request to get project admin's project Projects");
-//            result =  roleRepository.findAllRolesByProjectId(currentUser.getProject().getId()).stream()
-//                .map(roleMapper::roleToRoleDTO)
-//                .collect(Collectors.toCollection(LinkedList::new));
+            result = currentUser.getRoles().stream()
```
nivemaham: What is the use-case of this implementation?

dennyverbeeck: Good question, i noticed the project admin would not be able to get a list of roles so i fixed it and moved on. We used to have distinct pages to view roles, the pages still exist but the links inside the application do not, since looking at a role by itself is not very useful.

From what i can tell right now, it does not seem this method is even necessary anymore. I will investigate a bit further and create a separate issue for removing unused code related to roles.

PR #203, opened 2018-01-16
Add source demo data and enable snapshot releases to artifactory

nivemaham:
I have added some sample data for with a relevant source-type and added to dummy subject. This will come in handy to have the integration-tests, e2e tests based on dummy data, where we are using MP image on `dev` mode and testing for predefined subjects, sources and source-types. 
This should not affect the functionalities of MP on production mode. 
This PR also enables snapshot releases from MP and subprojects from `dev` branch


src/main/resources/config/liquibase/changelog/20180110164600_add_source_demo_data.xml
```diff
@@ -0,0 +1,29 @@
...
+  -->
+
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd
+                        http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
+
+    <changeSet id="20180110164600-0" author="nivethika@thehyve.nl">
```
dennyverbeeck: The `changeSet` tag should include the `context="dev, test"` attribute so it is only loaded for dev and test profiles.

PR #204, opened 2018-01-18
Release 0.3.3

nivemaham:
Release includes 
- Fixes to oauth-client-utils to create client specific instances of HttpClient
- `TokenException` and other security related exceptions to become explicitly catchable exceptions
- Check-style fixes
- Additional methods to validate radar tokens
- Add additional source, source-type data to be loaded on `dev` profile. 
- Enables snapshot releases to `artifactory` from `dev` branch


src/test/bash/test-docker-build-on-release-branch.sh
```diff
@@ -8,5 +8,6 @@ set -ev
 
 if [[ $TRAVIS_BRANCH == release-* ]]
 then
+  git checkout settings.gradle
```
dennyverbeeck: Perhaps better to remove this and move the line that modifies `settings.gradle` in the `.travis.yml` file to the end of the `script` section, because during deployment we will still want to publish the `managementportal-client` library.

nivemaham: Makes sense

PR #207, opened 2018-01-23
Travis tweaks

dennyverbeeck:
A number of tweaks to reduce Travis build times and eliminate unnecessary tests.

- No longer wait 60 seconds for app startup in `before_script`, now we check if the application is up right before we need it. If it's not up we wait for it to be up. This allows the checks to start running while the application is starting up in the background.
- Regular e2e tests do not run on a Travis `tag` build anymore, by the time we tag a commit, these tests will already have run and succeeded.
- Production e2e tests only run on release branches
- Docker builds only run on release branches

### DIFF
--- a/src/test/bash/run-prod-e2e.sh
+++ b/src/test/bash/run-prod-e2e.sh
@@ -20,7 +20,7 @@ then
   ./gradlew bootRepackage -Pprod buildDocker -x test
   docker-compose -f src/main/docker/app.yml up -d # spin up production mode application
   # wait for app to be up
-  ./../../../util/wait-for-app.sh
+  $TRAVIS_BUILD_DIR/util/wait-for-app.sh http://localhost:8080/managementportal/
   docker-compose -f src/main/docker/app.yml logs # show output of app startup
   yarn e2e # run e2e tests against production mode
   docker-compose -f src/main/docker/app.yml down -v # clean up containers and volumes
--- a/src/test/bash/run-prod-e2e.sh
+++ b/src/test/bash/run-prod-e2e.sh
@@ -10,23 +10,23 @@
 
 set -ev
 
-sed -i "s|new plugin.BaseHrefWebpackPlugin({ baseHref: '/' })|new plugin.BaseHrefWebpackPlugin({ baseHref: '/managementportal/' })|" webpack/webpack.dev.js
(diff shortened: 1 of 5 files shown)


src/test/bash/run-prod-e2e.sh
```diff
@@ -10,23 +10,23 @@
...
-echo ""
-echo "Application is up"
-docker-compose -f src/main/docker/app.yml logs # show output of app startup
-yarn e2e # run e2e tests against production mode
-docker-compose -f src/main/docker/app.yml down -v # clean up containers and volumes
-git checkout src/test/javascript/protractor.conf.js
-git checkout webpack/webpack.prod.js
-git checkout src/main/resources/config/application-prod.yml
+# only run on the release branch's push and pull_request events
+if [[ $TRAVIS_BRANCH == release-* ]] || [[ $TRAVIS_PULL_REQUEST_BRANCH == release-* ]]
```
blootsvoets: Perhaps run on the `master` branch as well, to see if there are any strange merge issues?

util/wait-for-app.sh
```diff
@@ -0,0 +1,19 @@
...
+else
+  APP_URL=$1
+fi
+
+# wait for app to start up
+echo "Waiting for application startup"
+until curl -s $APP_URL > /dev/null
+do
+  echo -n "." # waiting
+  sleep 2
```
blootsvoets: perhaps a timeout here would be good? e.g. maximum 3 minutes?

src/test/bash/run-prod-e2e.sh
dennyverbeeck: The `TRAVIS_PULL_REQUEST_BRANCH` environment variable is set for builds on the merge result, with its value set to the name of the source branch, so by the time the code gets in to master it should have already been tested. I'm not opposed to running on master again though if you feel it would bring added value. What do you think?

util/wait-for-app.sh
dennyverbeeck: Good point, will add it!

src/test/bash/run-prod-e2e.sh
dennyverbeeck: On the other hand, using `if [[ $TRAVIS_BRANCH == release-* ]] || [[ $TRAVIS_BRANCH == master ]]` will trigger on the PR build regardless of the source branch. Our current flow only merges into master from release branches, but if that would change in the future, then using this expression would mean we don't need to change this script. I'll update it!

PR #217, opened 2018-02-01
Require a valid token instead of trusting client_credentials

blootsvoets:
Not all clients necessarily have all permissions to do operations on all projects/subjects. Why was this exemption here in the first place?

### DIFF
--- a/radar-auth/src/main/java/org/radarcns/auth/token/AbstractRadarToken.java
+++ b/radar-auth/src/main/java/org/radarcns/auth/token/AbstractRadarToken.java
@@ -17,7 +17,6 @@ public abstract class AbstractRadarToken implements RadarToken {
     @Override
     public boolean hasPermission(Permission permission) {
         return hasScope(permission.scopeName());
-
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


blootsvoets:
Looking at the failing test, I'm wondering why the authority is checked at all in `RadarToken.hasPermission()`. It's obvious for anything related to a project or subject, but not for a general permission. Apparently MP thought that a user was allowed a certain scope, so then why check it against a role as well?

dennyverbeeck:
I think the original code was correct.

There is a difference between scope and authority. A scope is given to a token, while an authority is given to a user. A client is considered to have a permission if and only if

1. Their token has the correct scope
2. Their user has the correct authority
3. If it's a client credentials token, there is no user, therefore in that case there is no authority to check

Otherwise put, they have the permission if they `have the scope` AND (`is a client credentials token` OR `user has the correct authority`).

In `hasPermission()` we still need `hasAuthorityForPermission()` call. Scopes are part of the token, it is only saying "The user allowed this application to do this action on their behalf", which is different from "This application is allowed to do this action". Currently a user can authorize an application for any scope they wish, even if they don't have the appropriate authority. This could be seen as a disadvantage since we then need to check scope and authority always, but it is still secure since both scope and authority are always checked. There is also an advantage: an application can request and be granted a scope like PROJECT.UPDATE for a user that has not an admin for any project, if then the user later becomes the project admin for a project, this application will immediately be able to use that scope.

Hopefully this clarifies things, i'd be happy to continue discussion if needed :smile:

blootsvoets:
Thanks for the clarification. I understand it now.

PR #235, opened 2018-03-12
Inactive roles for subjects

### LINKED ISSUE #218: Do not remove ROLE_PARTICIPANT but make it inactive
Instead of removing the `ROLE_PARTICIPANT` role whenever a user is reassigned, it should be set to inactive. That way the REST API has a way of knowing whether a user was ever assigned to a given project. The implementation could have two parts:
1. the JWT only contains roles that are active, so participants will no longer be able to submit data for an inactive project.
2. the API exposes inactive participants, something like:
`GET /projects/radartest/participants`
```json
{
   "participants": [...],
   "inactiveParticipants": [...]
}
```

nivemaham:
Closes #218


radar-auth/src/main/java/org/radarcns/auth/authorization/Permissions.java
```diff
@@ -115,5 +116,10 @@ private static void initPermissions() {
         Arrays.asList(Permission.SUBJECT_READ, Permission.SUBJECT_UPDATE,
                 Permission.MEASUREMENT_CREATE, Permission.MEASUREMENT_READ).stream()
                         .forEach(p -> PERMISSION_MATRIX.get(p).add(PARTICIPANT));
+
+        /* Inactive participant */
+        // Can update and read own data and can read and write own measurements
```
dennyverbeeck: Small mistake here in comment, can not read or write measurements

src/main/java/org/radarcns/management/security/ClaimsTokenEnhancer.java
```diff
@@ -59,7 +60,9 @@ public OAuth2AccessToken enhance(OAuth2AccessToken accessToken,
             Optional<User> optUser = userRepository.findOneByLogin(userName);
             if (optUser.isPresent()) {
                 List<String> roles = optUser.get().getRoles().stream()
-                        .filter(role -> Objects.nonNull(role.getProject()))
+                        .filter(role -> Objects.nonNull(role.getProject())
+                                && !AuthoritiesConstants.INACTIVE_PARTICIPANT.equals(role
```
dennyverbeeck: Shouldn't we include the INACTIVE_PARTICIPANT role in the JWT anyway? If the token does not contain the role, then the user will not be able to get access to old project info based on this role.

src/main/java/org/radarcns/management/service/SubjectService.java
```diff
@@ -152,18 +156,34 @@ public SubjectDTO updateSubject(SubjectDTO subjectDto) {
             source.setAssigned(true);
         }
         // update participant role
-        Set<Role> managedRoles = subject.getUser().getRoles().stream()
-                .filter(r -> !AuthoritiesConstants.PARTICIPANT.equals(r.getAuthority().getName()))
-                .collect(Collectors.toSet());
-        managedRoles.add(getProjectParticipantRole(subjectDto.getProject()));
+        Set<Role> managedRoles = updateParticipantRoles(subject, subjectDto);
```
dennyverbeeck: Minor thing: I was a bit confused until i realized `subject` refers to the 'old' subject and `subjectDto` refers to the updated info. Perhaps the variable names could be `oldSubject` and `newSubject` or something along those lines.

src/main/java/org/radarcns/management/web/rest/SubjectResource.java
```diff
@@ -220,26 +222,54 @@
...
             throws NotAuthorizedException {
         checkPermission(getJWT(servletRequest), SUBJECT_READ);
         log.debug("ProjectName {} and external {}", projectName, externalId);
+        // if not specified do not include inactive patients
+        boolean withInactive = withInactiveParticipants != null ? withInactiveParticipants : false;
         if (projectName != null && externalId != null) {
-            Optional<Subject> subject = subjectRepository
-                    .findOneByProjectNameAndExternalId(projectName, externalId);
+            Optional<Subject> subject = Optional.empty();
+            if (!withInactive) {
```
dennyverbeeck: Also a minor thing: perhaps you can do a single check at the start of the method to create the correct list of authorities (either with or without `INACTIVE_PARTICIPANT`), and then use that list in the rest of the method, then you can remove the repeated checks on `withInactive`.

src/main/resources/config/liquibase/changelog/20180306165800_add_active_column_to_role.xml
```diff
@@ -0,0 +1,24 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd
+                        http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
+
+    <changeSet id="20180306165800-0" author="nivethika@thehyve.nl" >
+      <addColumn tableName="radar_role">
+        <column name="active_role" type="boolean" defaultValueBoolean="true"/>
+      </addColumn>
+    </changeSet>
+</databaseChangeLog>
```
dennyverbeeck: Do we still need this file? It's not referenced in master.xml.

src/main/java/org/radarcns/management/security/ClaimsTokenEnhancer.java
nivemaham: I don't think so. I would consider a token to have information that could only allow them to get certain access, not the data from history that is not valid anymore.

src/main/java/org/radarcns/management/web/rest/SubjectResource.java
nivemaham: indeed. That was the idea. :( I missed it

src/main/resources/config/liquibase/changelog/20180306165800_add_active_column_to_role.xml
nivemaham: Nope. we don't.

Issue #193, opened 2018-01-09
Remove projectAdmin field from Project domain class


dennyverbeeck:
The Project domain class still holds a dedicated field for project admin. This can be removed as project admins can be assigned with our roles model.

dennyverbeeck:
Pushing this into milestone v0.4.0, as it will be a DB schema change, and it will affect the audit logging #261 .

dennyverbeeck:
Closed by #261

PR #261, opened 2018-04-27
Enable envers

dennyverbeeck:
This PR enables full auditing with Hibernate Envers. There are still items to be completed, but the base functionality is there. I wanted to PR this work now, since I'm leaving for holiday, so it gives people some time to review.

Changes:
- Add `/revisions` API for querying general information about revisions
- Add `/subjects/revisions` API for querying subject revisions
- Change Subject-Source from many-to-many to one-to-many relationship
- Remove `projectAdmin` field from `Project` (#193)

Todos:
- Entities other than Subject still need to get their revision API and UI elements.
- Tests still need to be written, Java as well as e2e tests.
- Update documentation


src/main/java/org/radarcns/management/web/rest/SubjectResource.java
```diff
@@ -289,6 +296,53 @@
...
+    public ResponseEntity<List<RevisionDTO>> getSubjectRevisions(
+            @ApiParam Pageable pageable, @PathVariable String login) throws NotAuthorizedException {
+        log.debug("REST request to get revisions for Subject : {}", login);
+        SubjectDTO subject = subjectService.getLatestRevision(login);
+        Page<RevisionDTO> page = revisionService.getRevisionsForEntity(pageable,
+                subjectMapper.subjectDTOToSubject(subject));
+        SubjectDTO blank = new SubjectDTO();
+        RadarToken token = getJWT(servletRequest);
+        // iterate the revisions, if we don't have read permission for the subject in a given
+        // revision, change the subject with a 'blank' subject so no information gets leaked
```
nivemaham: I would rather have another API to request revisions of subjects based on project, having one API expose partial information depends on the permission. For wrong project request we should give 401 instead of blank subject to maintain the consistency.

src/main/webapp/app/entities/revision/revision.route.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const revisionRoute: Routes = [
+    {
+        path: 'revisions',
+        component: RevisionComponent,
+        resolve: {
+            'pagingParams': ResolvePagingParams
+        },
+        data: {
+            authorities: [SYSTEM_ADMIN, PROJECT_ADMIN],
```
nivemaham: Are project admins allowed to see revisions?

src/main/webapp/app/layouts/navbar/navbar.component.html
```diff
@@ -127,6 +127,12 @@
                             <span jhiTranslate="global.menu.admin.audits">Audits</span>
                         </a>
                     </li>
+                    <li *jhiHasAnyAuthority="['ROLE_SYS_ADMIN']">
+                        <a class="dropdown-item" routerLink="revisions" routerLinkActive="active" (click)="collapseNavbar()">
```
nivemaham: The auth-guard is configured to allow PROJECT_ADMIN and SYS_ADMIN and UI is allowed for SYSTEM_ADMIN only

src/main/webapp/app/shared/subject/subject-detail.component.html
```diff
@@ -36,14 +36,6 @@ <h2><span jhiTranslate="managementPortalApp.subject.detail.title">Subject</span>
                 <a [routerLink]="['/source', sources?.sourceName ]">{{sources.sourceTypeModel}}: {{sources.sourceName}}</a>{{last ? '' : ', '}}
             </span>
         </dd>
-        <dt><span jhiTranslate="managementPortalApp.subject.createdBy">Created By</span></dt>
-        <dd>{{subject.createdBy}}</dd>
-        <dt><span jhiTranslate="managementPortalApp.subject.createdDate">Created Date</span></dt>
-        <dd>{{subject.createdDate | date:'dd/MM/yy HH:mm' }}</dd>
-        <dt><span jhiTranslate="managementPortalApp.subject.lastModifiedBy">Last Modified By</span></dt>
```
nivemaham: Why do we remove these from UI?

gradle/liquibase.gradle
```diff
@@ -40,11 +40,15 @@ task liquibaseDiffChangelog(dependsOn: compileJava, type: JavaExec) {
...
+    args "--referenceUrl=hibernate:spring:org.radarcns.management.domain?dialect=org.hibernate.dialect.PostgreSQLDialect&hibernate.physical_naming_strategy=org.springframework.boot.orm.jpa.hibernate.SpringPhysicalNamingStrategy&hibernate.implicit_naming_strategy=org.springframework.boot.orm.jpa.hibernate.SpringImplicitNamingStrategy"
+    // comment out line above and uncomment lines below if you want to compare against another postgres instance
+    //args "--referenceUrl=jdbc:postgresql://localhost:5433/managementportal"
+    //args "--referenceUsername=postgres"
+    //args "--referencePassword="
+    args "--username=postgres"
     args "--password="
-    args "--url="
-    args "--driver="
+    args "--url=jdbc:postgresql://localhost:5432/managementportal"
```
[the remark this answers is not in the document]
dennyverbeeck: Ok good point, i'll comment it out and add a comment with some explanation

src/main/java/org/radarcns/management/domain/Source.java
```diff
@@ -1,46 +1,49 @@
...
 @Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)
-public class Source extends AbstractAuditingEntity implements Serializable {
+@EntityListeners({AbstractEntityListener.class})
+public class Source extends AbstractEntity implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
     @Id
-    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "sequenceGenerator")
+    @GeneratedValue(strategy = GenerationType.AUTO, generator = "sequenceGenerator")
```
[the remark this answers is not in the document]
dennyverbeeck: Ok, we change this back to `GenerationType.SEQUENCE`

src/main/java/org/radarcns/management/domain/audit/CustomRevisionEntity.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @RevisionTimestamp
+    private Date timestamp;
+
+    private String auditor;
+
+    @ElementCollection(fetch = FetchType.EAGER)
+    @JoinTable(name = "REVCHANGES", joinColumns = @JoinColumn(name = "REV"))
+    @Column(name = "ENTITYNAME")
+    @Fetch(FetchMode.JOIN)
```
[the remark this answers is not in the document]
dennyverbeeck: Not sure about it, i'll try to find out more info about it

src/main/java/org/radarcns/management/domain/support/AbstractEntityListener.java
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * Event listener to log a persist event.
+     *
+     * @param entity the entity that is persisted
+     */
+    @PostPersist
+    public void publishPersistEvent(AbstractEntity entity) {
+        AutowireHelper.autowire(this, springSecurityAuditorAware);
+        log.info(TEMPLATE, ENTITY_CREATED, springSecurityAuditorAware.getCurrentAuditor(),
```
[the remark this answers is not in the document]
dennyverbeeck: Exactly, so this class only serves to log changes to the console, so admins can have a quick look at the log output to see what's happening in MP.

src/main/java/org/radarcns/management/service/SubjectService.java
```diff
@@ -146,35 +155,35 @@ public SubjectDTO updateSubject(SubjectDTO newSubjectDto) {
...
+        sourceRepository.save(sourcesToUpdate);
         // update participant role
         Set<Role> managedRoles = updateParticipantRoles(subjectFromDb, newSubjectDto);
-        subjectFromDb.getUser().setRoles(managedRoles);
-        subjectFromDb = subjectRepository.save(subjectFromDb);
-
-        return subjectMapper.subjectToSubjectDTO(subjectFromDb);
+        // seems silly to do this check, but if we don't we'll always get a user update in the
+        // revision log, even if the roles did not change, and even if the new set equals the old
+        // set, not sure why envers does it this way
```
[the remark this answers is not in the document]
dennyverbeeck: Hmm true, better to keep the code a bit cleaner since indeed they are in a OneToOne relation so it does make sense.

src/main/java/org/radarcns/management/service/SubjectService.java
```diff
@@ -357,4 +369,65 @@ public void deleteSubject(String login) {
...
+
+    /**
+     * Get latest known revision of a subject with the given login.
+     *
+     * @param login the login of the subject
+     * @return the latest revision for that subject
+     * @throws CustomNotFoundException if no subject was found with the given login
+     */
+    public SubjectDTO getLatestRevision(String login) throws CustomNotFoundException {
+        UserDTO user = (UserDTO) revisionService.getLatestRevisionForEntity(User.class,
```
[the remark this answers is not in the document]
dennyverbeeck: I'll check it again to make sure, but i believe it was not possible to look for a property of a linked object when the object itself was deleted. I.e. look for `user.login` if the subject i'm looking for is deleted. This method should be able to return the latest known revision, even if that revision was a deletion.

src/main/java/org/radarcns/management/web/rest/SubjectResource.java
dennyverbeeck: The problem with such an API is that a client would have to iterate over all known projects to get a full history for a subject. 

I was thinking of the scenario where a subject is moved to another project, perhaps even multiple times. If i'm a project admin for 2 out of the 4 projects a subject had been a member of, i should be able to see the changes to the subject, only when the subject was registered to those two projects i have access to. For changes that occurred when the subject was registered to the other two projects, the rationale is to show there has been a change, so still show the revision, but not show the contents, that is where the 'blank' subject comes in. Otherwise the history this method returns is a false one.

Of course this discussion becomes pointless if a project admin can not access any historical data, since sys admins have access to all information anyway. We'll ask the wider team what roles should be able to see historical data.

src/main/java/org/radarcns/management/domain/audit/CustomRevisionEntity.java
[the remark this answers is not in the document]
dennyverbeeck: So, `FetchType.EAGER` instructs the persistence layer *when* to fetch associated  entities, while `FetchMode.JOIN` instructs it *how* to do it. Besides `JOIN` there are options `SELECT` (i.e. do a new select query for each instance in the collection) and `SUBSELECT` which will fetch all elements in the collection with a single secondary select statement.

src/test/javascript/e2e/scenarios/discontinue-subject-should-unassign-source.spec.ts
```diff
@@ -120,19 +124,6 @@ describe('Discontinued subject should unassign sources', () => {
         });
     });
 
-    it('should be able to delete a source', function () {
-        element.all(by.linkText(sourceName))
-            .all(by.xpath('ancestor::tr'))
-            .all(by.cssContainingText('button', 'Delete'))
-            .click().then(() => {
-                element(by.css('.modal-footer button.btn-danger')).click().then(() => {
```
nivemaham: we should not be able to delete a source that was assigned to subject. Maybe we can change this check to test that?

src/main/java/org/radarcns/management/web/rest/SubjectResource.java
```diff
@@ -377,12 +427,13 @@
         Long sourceTypeId = sourceDto.getSourceTypeId();
         if (sourceTypeId == null) {
             // check if combination (producer, model, version) is present
-            final String msg = "error.sourceNotFound";
             try {
-                String producer = Objects.requireNonNull(sourceDto.getSourceTypeProducer(), msg);
-                String model = Objects.requireNonNull(sourceDto.getSourceTypeModel(), msg);
+                String producer = Objects.requireNonNull(sourceDto.getSourceTypeProducer(),
+                        ErrorConstants.ERR_SOURCE_NOT_FOUND);
+                String model = Objects.requireNonNull(sourceDto.getSourceTypeModel(),
```
nivemaham: should it be `SOURCE_TYPE_NOT_FOUND`?

README.md
```diff
@@ -71,7 +71,7 @@ memory database and ManagementPortal.
 
 
 
-The docker image can be pulled by running `docker pull radarcns/management-portal:0.3.8
+The docker image can be pulled by running `docker pull radarbase/management-portal:0.3.8
```
nivemaham: and the missing ` and the end of the line too. pls :)

dennyverbeeck: Nice 😄  done!

src/main/java/org/radarcns/management/web/rest/SubjectResource.java
dennyverbeeck: Actually these checks should throw a validation error if any of these are null, because that would be a bad request from the client. Will change.

src/test/javascript/e2e/scenarios/discontinue-subject-should-unassign-source.spec.ts
dennyverbeeck: Makes sense!

PR #278, opened 2018-06-29
Resend activation link

### LINKED ISSUE #274: Add ability to resend an activation link
Right now if a user misses the 24hr deadline to activate their account, we have to delete and re-create the user to send the activation link. 
There should be an option to resend it in this case.

nivemaham:
Enables resending activation link by creating new reset-key and reseting expiry.

Fixes #274


nivemaham:
The reset key is reset when we send the activation link again. so they cannot use the old one. Just the latest one works

PR #282, opened 2018-07-05
Attribute for human readable project name

### LINKED ISSUE #273: Human readable projectName as an additional attribute for project and show that as default value in UI if exists
Related to https://github.com/RADAR-base/RADAR-RestApi/issues/98

nivemaham:
- Use HashMap instead of AttributeDTO in projectDTO
- Adds a field to represent humanReadableProjectName
- UI adds another required field to input Human readable project name and the existing one is referred as project id.
- a Check-box is added to UI to use the project-id as the value for human readable project name

Fixed #273


src/main/java/org/radarcns/management/service/mapper/decorator/ProjectMapperDecorator.java
```diff
@@ -29,15 +28,12 @@ public ProjectDTO projectToProjectDTO(Project project) {
...
-        Set<AttributeMapDTO> attributeMapDtoList = new HashSet<>();
-        if (project.getAttributes() != null) {
-            for (Entry<String, String> entry : project.getAttributes().entrySet()) {
-                AttributeMapDTO attributeMapDto = new AttributeMapDTO();
-                attributeMapDto.setKey(entry.getKey());
-                attributeMapDto.setValue(entry.getValue());
-                attributeMapDtoList.add(attributeMapDto);
-            }
-            dto.setAttributes(attributeMapDtoList);
+        List<String> humanReadablename = project.getAttributes().entrySet().stream()
```
blootsvoets: Why not just an attributes.get?

PR #285, opened 2018-07-10
Update source attributes

### LINKED ISSUE #196: Update source attributes
It would be nice to have a mechanism to re-register sources or at least update their attributes. For example, firmware version, Android version, etc. Right now the pRMT is using the `POST api/subjects/{login}/sources` query, but that will give `409 Conflict` on double posts. I'd expect a `POST api/subjects/{login}/sources/{sourceId}` call that takes a subset of the Source properties and updates those.

There is a `PUT /api/sources` call, but at least the pRMT does not have access to that. Generally, the `PUT` request method is meant to replace a resource, but I only want to update the existing source.

nivemaham:
- Adds a `POST` request to update attributes of a source added to subject. 
Fixes #196

### DIFF
--- a/src/main/java/org/radarcns/management/service/SourceService.java
+++ b/src/main/java/org/radarcns/management/service/SourceService.java
@@ -1,9 +1,6 @@
 package org.radarcns.management.service;
 
-import java.net.URI;
-import java.net.URISyntaxException;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


src/main/java/org/radarcns/management/service/SourceService.java
```diff
@@ -131,4 +135,32 @@ public void delete(Long id) {
...
+     * attributes only.
+     * @param sourceToUpdate source fetched from database
+     * @param attributes value to update
+     * @return Updated {@link MinimalSourceDetailsDTO} of source
+     */
+    public MinimalSourceDetailsDTO safeUpdate(Source sourceToUpdate,
+            Map<String, String> attributes) {
+
+        // update source attributes
+        Map<String, String> mergedValues = Stream.of(sourceToUpdate.getAttributes(), attributes)
```
blootsvoets: why not `sourceToUpdate.getAttributes().putAll(attributes)`? Or
```java
Map<String, String> updatedAttributes = new HashMap<>();
updatedAttributes.putAll(sourceToUpdate.getAttributes());
updatedAttributes.putAll(attributes);
```

PR #286, opened 2018-07-12
Fix source transfer

### LINKED ISSUE #239: When a Source is transferred to new project, it's source-type should remain compatible.
When a Source is transferred to new project, it's source-type should remain compatible , e.g. Android-toAndroid, Empatica-to-Empatica, but not Empatica-Android.

nivemaham:
- Adds validation for source update and source transfer occasions.
- A source cannot be transfered to another project if it is assigned
- A source cannot be transfered to a project which doesn't have compatible source-type.

Fixes #239

### DIFF
--- a/src/main/java/org/radarcns/management/web/rest/SourceResource.java
+++ b/src/main/java/org/radarcns/management/web/rest/SourceResource.java
@@ -18,6 +18,8 @@ import javax.servlet.http.HttpServletRequest;
 import javax.validation.Valid;
 import org.radarcns.auth.config.Constants;
 import org.radarcns.auth.exception.NotAuthorizedException;
+import org.radarcns.management.domain.SourceType;
+import org.radarcns.management.repository.ProjectRepository;
 import org.radarcns.management.repository.SourceRepository;
 import org.radarcns.management.service.SourceService;
 import org.radarcns.management.service.dto.SourceDTO;
@@ -59,6 +61,9 @@ public class SourceResource {
 
     @Autowired
     private HttpServletRequest servletRequest;
+
+    @Autowired
+    private ProjectRepository projectRepository;
 
     /**
      * POST  /sources : Create a new source.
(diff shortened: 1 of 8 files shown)


src/main/java/org/radarcns/management/web/rest/SourceResource.java
```diff
@@ -114,6 +119,38 @@
...
+            if  (!sourceType.isPresent()) {
+                return ResponseEntity.badRequest().headers(HeaderUtil
+                    .createFailureAlert(ENTITY_NAME,
+                    "invalidTransfer", "Cannot transfer a source to a project which doesn't "
+                        + "have compatible source-type")).build();
+            }
+            // set old source-type, ensures compatibility
+            sourceDto.setSourceType(sourceToUpdate.getSourceType());
+
+        }
```
dennyverbeeck: I think we should move all this logic into `SourceService`, **including** the call to `save()`. This would make the code more re-usable, plus it would make the whole update transactional.

nivemaham: Indeed. That was my preferred method. But the problem is mainly due to the inconsistent way of error handling. We have two ways of error handling. One is by throwing exception and have the `ExceptionTranslator` to translate this into body. However, it deviates from the original method that was followed in the FrontEnd app. It expects headers with custom parameters to show proper messages to the user. So I decided to stick to responding with headers, because this action mainly involves user interaction and proper message on the front end seems essential. I didn't want to return `ResponseEntity` from `SourceService` either. It sounded far from `SourceService` logics. So I chose this way.

dennyverbeeck: Hmm, i thought the frontend could handle also error messages in the body, but i might be mistaken. If not we can modify the `ExceptionTranslator` to set the headers, and use exceptions everywhere.

Ok let's keep it here for now and revisit the error handling later, but i would feel safer with at least the `save()` being delegated to the service so it's transactional 😄

nivemaham: Yes. That is what I thought. So I will push the exceptionHandler changes too. I was thinking of doing it separately. But same PR doesn't hurt.
Save is done through `SourceService` which is Transactional. 
`  SourceDTO result = sourceService.save(sourceDto);`

nivemaham: I don't know why we have specific annotations for some methods in classes which are Transactional.

dennyverbeeck: > Save is done through SourceService which is Transactional. `SourceDTO result = sourceService.save(sourceDto);`

Sorry, of course, my bad!

> I don't know why we have specific annotations for some methods in classes which are Transactional.

Indeed, the method-level annotations can go away if there is a class-level one.

PR #295, opened 2018-07-26
Meta qr code

### LINKED ISSUE #275: Switch to meta-QR code
As discussed in our weekly meeting, we will no longer encode the refresh token in the QR code for pairing an app. Instead we will generate a unique URL where the information for pairing will be hosted, and we will encode this URL in the QR code instead. This will make QR codes have less data encoded in them, so they become easier to scan. Additionally, related to #262, this will allow us to push more information to the client without increasing the size of the QR code.

nivemaham:
**Move to the use of MetaQR code**
- Created a new entity called meta-token to store the token and some meta-data about it.
- The QR code is generated with a `URL` and a `tokenName` instead of token.
- A new REST endpoint is added to `GET` the token.
- A `tokenTimeout` should be set in the configs. Note: This determines for how long the token is valid. The currently the REST endpoint is not protected. So the shorted the timeout is better for security. 
- A cron job is introduced to clean stale meta-tokens ( Expired and fetched)

Fixes #275


radar-auth/src/main/java/org/radarcns/auth/config/Constants.java
```diff
@@ -7,6 +7,7 @@
 
     //Regex for acceptable logins
     public static final String ENTITY_ID_REGEX = "^[_'.@A-Za-z0-9- ]*$";
+    public static final String TOKEN_NAME_REGEX = "^[A-Za-z0-9]*$";
```
blootsvoets: is a dash `-` also acceptable?

src/main/java/org/radarcns/management/domain/MetaToken.java
```diff
@@ -0,0 +1,135 @@
...
+    private String tokenName;
+
+    @Column(name = "token", length = 2000)
+    private String token;
+
+    @Column(name = "fetched", nullable = false)
+    private Boolean fetched;
+
+    @Column(name = "expiry_date")
+    private ZonedDateTime expiryDate = null;
```
blootsvoets: Why does this need a time zone? Can it be an `Instant` instead for timezone-independent expiry instant?

src/main/java/org/radarcns/management/domain/MetaToken.java
```diff
@@ -0,0 +1,135 @@
...
+@Table(name = "radar_meta_token")
+@Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)
+@EntityListeners({AbstractEntityListener.class})
+public class MetaToken extends AbstractEntity {
+
+
+    //https://math.stackexchange.com/questions/889538/
+    // probability-of-collision-with-randomly-generated-id
+    // Current length of tokenName is 12. If we think there might be collision we can increase
+    // the length.
```
blootsvoets: There are almost 6 bits of entropy per character (A-Za-z0-9 space size is 62 is close to 2^6). So there are 72 bits of entropy for the entire string. That is almost 5000 * billion * billion possibilities. For the number of users we're looking at, it should be enough for security. However, we still have to do collision detection.

src/main/java/org/radarcns/management/web/rest/OAuthClientsResource.java
```diff
@@ -263,8 +274,22 @@
...
+                .fetched(false);
+        metaToken = metaTokenService.save(metaToken);
+        ClientPairInfoDTO cpi = null;
+        if (metaToken.getId() != null && metaToken.getTokenName() != null) {
+            // get base url from settings
+            String baseUrl = managementPortalProperties.getMail().getBaseUrl();
+            // format base url
+            baseUrl = baseUrl.endsWith("/") ? baseUrl : baseUrl + "/";
+            // create complete uri string
+            String url = baseUrl + "#" + ResourceUriService.getUri(metaToken).getPath();
```
blootsvoets: remove the hash :-)

radar-auth/src/main/java/org/radarcns/auth/config/Constants.java
nivemaham: No it is not.

src/main/java/org/radarcns/management/domain/MetaToken.java
nivemaham: I made it ZonedDateTime just to be consistant with other time based properties we have in other entities. If Instant is preferred, it can be done

src/main/java/org/radarcns/management/web/rest/OAuthClientsResource.java
nivemaham: My bad. my bad :man_facepalming:

src/main/java/org/radarcns/management/domain/MetaToken.java
blootsvoets: For user-local or server-local time notation (e.g. showing something on the screen, logging user input), adding a time zone makes sense. For user-independent time stamps (something should just happen at a certain time) I think Instant is probably less error-prone (because it avoids mapping to the correct time zone).

src/main/java/org/radarcns/management/web/rest/MetaTokenResource.java
```diff
@@ -39,7 +40,14 @@
...
-        return ResponseEntity.ok().body(metaTokenService.fetchToken(tokenName));
+        TokenDTO retrievedToken = metaTokenService.fetchToken(tokenName);
+
+        try {
+            TimeUnit.SECONDS.sleep(1);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        }
+
+        return ResponseEntity.ok().body(retrievedToken);
```
blootsvoets: With 404 if not found? The sleep should also be present for 404s

src/main/java/org/radarcns/management/service/MetaTokenService.java
```diff
@@ -0,0 +1,94 @@
...
+        } else {
+            throw new CustomParameterizedException("invalidRequest", "token is already fetched. "
+                + "Invalid request");
+        }
+
+
+    }
+
+    @Transactional(readOnly = true)
+    public MetaToken getToken(String tokenName) {
```
blootsvoets: Perhaps this can be Optional to beter support sleeping on not found. Alternatively, sleep in the current method and not in the API call, but that will make all internal calls and this transaction slower.

src/main/java/org/radarcns/management/web/rest/MetaTokenResource.java
```diff
@@ -38,16 +42,16 @@
...
-
         try {
-            TimeUnit.SECONDS.sleep(1);
+            // sleep for 5 seconds
+            TimeUnit.SECONDS.sleep(5);
+            TokenDTO retrievedToken = metaTokenService.fetchToken(tokenName);
+            return ResponseEntity.ok().body(retrievedToken);
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
+            return ResponseEntity.unprocessableEntity().build();
```
blootsvoets: I’d expect HTTP status code 500 here, as unexpected server state.

src/main/java/org/radarcns/management/service/MetaTokenService.java
```diff
@@ -55,16 +56,17 @@ public MetaToken save(MetaToken metaToken) {
...
-    public TokenDTO fetchToken(String tokenName) throws CustomNotFoundException,
-            MalformedURLException {
+    public TokenDTO fetchToken(String tokenName) throws
+            MalformedURLException, CustomNotFoundException {
         log.debug("Request to get Token : {}", tokenName);
         MetaToken fetchedToken = getToken(tokenName);
 
-        if (!fetchedToken.isFetched()) {
+        // process the response if the token is not fetched or not expired
+        if (!fetchedToken.isFetched() && Instant.now().isBefore(fetchedToken.getExpiryDate())) {
```
blootsvoets: If expired, remove from database as well? Or is there a separate process for removing expired tokens?

nivemaham: My plan was to have a scheduled job to remove expired and fetched tokens

src/main/java/org/radarcns/management/web/rest/MetaTokenResource.java
nivemaham: Sure. Will change the status code

src/main/java/org/radarcns/management/service/MetaTokenService.java
```diff
@@ -0,0 +1,135 @@
...
+                ErrorConstants.ERR_TOKEN_NOT_FOUND,
+                Collections.singletonMap("tokenName", tokenName));
+        }
+    }
+
+    /**
+     * Builds a unique meta-token instance, by checking for token-name collision.
+     * @return an unique token
+     */
+    public MetaToken buildUniqueToken() {
```
blootsvoets: Because the search for a unique token is not atomic with saving a token, this does not guarantee a unique token. Flow: process 1 creates token, process 2 creates identical token, both are stored one after the other. The second `save` call should give a conflict from the DB, because there is a unique constraint. Ideally, buildUniqueToken would also store the token and solve the conflict by catching an exception and then choosing a new token name, instead of searching. Or catch it one layer up, in the `/oauth-clients/pair` request handling. This is also faster than searching the repository, because it is an exceptional situation. So in most situations, this check is going to return false, but it still costs some overhead.

nivemaham: Ok. Would creating identical token includes the `id` of the token or should I skip it so it focuses on the `tokenName` unique constraint?

blootsvoets: The ID would be autogenerated, right? So it will always be unique. But the tokenName is generated by this code, so we need to resolve conflicts there as well.

nivemaham: Yes. ID is auto-generated.

src/main/java/org/radarcns/management/service/MetaTokenService.java
```diff
@@ -103,20 +105,26 @@ public MetaToken getToken(String tokenName) throws CustomNotFoundException {
...
+     * If a collision is detection, we try to save the token with a new tokenName
      * @return an unique token
      */
-    public MetaToken buildUniqueToken() {
-        MetaToken token = new MetaToken();
-
-        while (metaTokenRepository.findOneByTokenName(token.getTokenName()).isPresent()) {
-            token.tokenName(RandomStringUtils.randomAlphanumeric(SHORT_ID_LENGTH));
+    public MetaToken saveUniqueToken(String token, Boolean fetched, Instant expiryTime) {
+        MetaToken metaToken = new MetaToken()
```
blootsvoets: still need to add the `.tokenName(RandomStringUtils.randomAlphanumeric(SHORT_ID_LENGTH))` builder method

nivemaham: The the MetaToken constructor assigns `RandomStringUtils.randomAlphanumeric(SHORT_ID_LENGTH)` to tokenName. That is why I didn't have it explicitly.

PR #296, opened 2018-08-03
Harmonize exception handling

nivemaham:


src/main/java/org/radarcns/management/service/SubjectService.java
```diff
@@ -324,19 +301,52 @@ public MinimalSourceDetailsDTO assignOrUpdateSource(Subject subject, SourceType
...
             Map<String, String> errorParams = new HashMap<>();
-            errorParams.put("message", "Cannot find assigned source with sourceId or a source of "
-                    + "sourceType with the specified producer and model is already registered "
-                    + "for subject login ");
             errorParams.put("producer", sourceType.getProducer());
             errorParams.put("model", sourceType.getModel());
             errorParams.put("subject-id", subject.getUser().getLogin());
             errorParams.put("sourceId", sourceRegistrationDto.getSourceId().toString());
-            throw new CustomParameterizedException("InvalidRequest", errorParams);
+            throw new BadRequestException("Cannot find assigned source with sourceId or a source "
```
blootsvoets: Is this not a conflictexception? E.g. http status code 409?

src/main/java/org/radarcns/management/service/SubjectService.java
```diff
@@ -406,21 +416,38 @@ public SubjectDTO findRevision(String login, Integer revision) throws CustomNotF
...
     }
+
+
+    /**
+     * Finds {@link Subject} from databased from login provided.
+     * @param login of subject to look for.
+     * @return {@link Subject} loaded.
+     */
+    public Subject findOneByLogin(String login) {
+        Optional<Subject> subject = subjectRepository.findOneWithEagerBySubjectLogin(login);
```
blootsvoets: Use `Optional#orElseThrow()` in these cases

src/main/java/org/radarcns/management/service/SubjectService.java
nivemaham: This cannot be a conflict request, since we did not get enough data to process this request. This is a situation where we cannot both create a new source, or update source.

src/main/java/org/radarcns/management/web/rest/SourceTypeResource.java
```diff
@@ -224,14 +226,13 @@
...
                 version);
         if (!projects.isEmpty()) {
-            throw new CustomParameterizedException(ErrorConstants.ERR_SOURCE_TYPE_IN_USE,
-                    projects.stream()
-                            .map(p -> p.getProjectName())
-                            .reduce((s1, s2) -> String.join(", ", s1, s2))
-                            .get()); // we know the list is not empty so calling get() is safe here
+            throw new InvalidRequestException(
+                // we know the list is not empty so calling get() is safe here
+                "Cannot delete a source-type that " + "is being used by project(s)", SOURCE_TYPE,
```
blootsvoets: Add back the list of projects here? The reduce(String.join).get() can be  replaced with collect(Collectors.joining)

src/main/java/org/radarcns/management/service/SubjectService.java
blootsvoets: Ok understood

Issue #297, opened 2018-08-06
Add privacy policy url to metaQR code. 


nivemaham:
Adds `privacy policy url` to metaQR code. This should give project specific url if available or a general privacy policy url configured.

Issue #289, opened 2018-07-17
Admin password should be configurable via Environment variable


yatharthranjan:
The admin password should be configurable via an environment variable by default unless changed via the portal. Related to https://github.com/RADAR-base/RADAR-Docker/issues/137

nivemaham:
Why not just change the password from MP front-end?

yatharthranjan:
This is for securing the MP before that happens. If the change is done on the front end then this env var will have no effect. This also helps keeps all the configuration consistent in one place when installing the stack.
If this is not trivial to do in the MP, then yes we can leave it as it not a requirement but something good to have.

yatharthranjan:
Also there are more chance of someone not updating the password on the front end. But if it has to be configured as part of installation of the stack then there is definitely going to be better secured.

PR #298, opened 2018-08-07
Improvements and fixes

### LINKED ISSUE #291: Crash on radar-schemas 0.3.4
The management portal crashes when it uses catalog server with radar-schemas 0.3.4. I'm assuming this is because it does not understand the newly created `CONNECTOR` source type, and that doesn't have a vendor, model and version.

Ideally, the management portal should not crash when it encounters a wrong source type, but warn and ignore instead.

nivemaham:
- Adds `privacyPolicyUrl` property to MetaQR code response
- Ignores errors during source-type loading from catalog server with warning message.
- Adds configuration to override default admin password. 
Fixes #291 #297 #289


src/main/java/org/radarcns/management/web/rest/SubjectResource.java
```diff
@@ -441,18 +427,31 @@
                         sourceDto.getSourceTypeCatalogVersion()).getId();
 
         }
+
+        // check the subject id
+        Subject sub = subjectService.findOneByLogin(login);
+
+        // find the actively assigned project for this subject
+        Optional<Project> activeProject = sub.getActiveProject();
```
blootsvoets: use `orElseThrow`?

src/main/java/org/radarcns/management/domain/Subject.java
```diff
@@ -155,6 +158,26 @@ public void setAttributes(Map<String, String> attributes) {
...
+     * and the project that is related to that role is the active role.</p>
+     *
+     * @return {@link Project} currently active project of subject.
+     */
+    public Optional<Project> getActiveProject() {
+        Optional<Role> activeProjectRole = this.getUser().getRoles().stream()
+                .filter(r -> r.getAuthority().getName().equals(PARTICIPANT))
+                .findFirst();
+
+        if (activeProjectRole.isPresent()) {
```
blootsvoets: instead of if/else, you can do
```java
return activeProjectRole.map(Role::getProject);
```

blootsvoets: Or even do that one line up:
```
...
.findFirst()
.map(Role::getProject);
```

src/main/java/org/radarcns/management/service/OAuthClientService.java
```diff
@@ -201,15 +205,55 @@ public ClientPairInfoDTO createRefreshToken(Subject subject, String clientId)
...
+     */
+    private URL getPrivacyPolicyUrl(Subject subject) {
+
+        Optional<Project> project = subject.getActiveProject();
+
+        // load default url from config
+        String policyUrl = managementPortalProperties.getCommon().getPrivacyPolicyUrl();
+
+        // override from project specific url if available.
+        if (project.isPresent()) {
```
blootsvoets: the lines above can be shortened to:
```java
String policyUrl = subject.getActiveProject()
    .map(p -> p.getAttributes().get(PRIVACY_POLICY_URL))
    .filter(u -> u != null && !u.isEmpty())
    .orElse(managementPortalProperties.getCommon().getPrivacyPolicyUrl());
```

src/main/java/org/radarcns/management/service/OAuthClientService.java
```diff
@@ -201,15 +205,55 @@ public ClientPairInfoDTO createRefreshToken(Subject subject, String clientId)
...
+            if (projectPrivacyPolicyUrl != null && !projectPrivacyPolicyUrl.isEmpty()) {
+                policyUrl = projectPrivacyPolicyUrl;
+            }
+        }
+
+        try {
+            return new URL(policyUrl);
+        } catch (MalformedURLException e) {
+            Map<String, String> params = new HashMap<>();
+            params.put("Url" , policyUrl);
```
blootsvoets: lowercase `url`?

src/main/java/org/radarcns/management/service/mapper/decorator/SubjectMapperDecorator.java
```diff
@@ -39,12 +38,10 @@ public SubjectDTO subjectToSubjectDTO(Subject subject) {
...
         SubjectDTO dto = delegate.subjectToSubjectDTO(subject);
         dto.setStatus(getSubjectStatus(subject));
-        Optional<Role> role = subject.getUser().getRoles().stream()
-                .filter(r -> r.getAuthority().getName().equals(AuthoritiesConstants.PARTICIPANT))
-                .findFirst();
-
-        role.ifPresent(role1 ->
-                dto.setProject(projectMapper.projectToProjectDTO(role1.getProject())));
+        Optional<Project> activeProject = subject.getActiveProject();
+        if (activeProject.isPresent()) {
```
blootsvoets: why not `ifPresent(project -> dto.setProject(projectMapper.projectToProjectDTO(project)))`?

nivemaham: Pushed the changes. Thanks for the suggestions.

PR #306, opened 2018-08-13
Fix null pointer on config load

### LINKED ISSUE #304: YamlServerConfig throws NullPointerException on file not found
On line
https://github.com/RADAR-base/ManagementPortal/blob/5bca05baa55520b20a0b82d2d35ed5c1336161d5/radar-auth/src/main/java/org/radarcns/auth/config/YamlServerConfig.java#L60
if the `radar-is.yml` file is not found on the class path, the logging statement throws a `NullPointerException`. I would expect a `ConfigurationException` instead as stated in the Javadoc.

nivemaham:
Fixes #304


radar-auth/src/main/java/org/radarcns/auth/config/YamlServerConfig.java
```diff
@@ -40,39 +38,44 @@
      * @return The initialized configuration object based on the contents of the configuration file
      * @throws ConfigurationException If there is any problem loading the configuration
      */
-    public static YamlServerConfig readFromFileOrClasspath() {
+    public YamlServerConfig loadConfig() {
```
blootsvoets: Why is this? It makes sense for it to be a static function. Perhaps the stored `config` variable could be removed though, and it can always be loaded from scratch. Alternatively, it could use a singleton pattern, but just loading it every time makes more sense to me.

nivemaham: Codacy suggested to remove static singletons. or it should be synchronized. I will have a look again

blootsvoets: Indeed, and if the `config` variable is removed it doesn't have to be synchronised, then it is completely stateless.

nivemaham: That would make the file to be loaded every time we call the method. Is that what we expect?

radar-auth/src/main/java/org/radarcns/auth/config/YamlServerConfig.java
```diff
@@ -38,15 +37,7 @@
...
-    public YamlServerConfig loadConfig() {
-        if (config == null) {
-            config = readFromFileOrClasspath();
-        }
-        return config;
-    }
-
-
-    private YamlServerConfig readFromFileOrClasspath() {
+    public static synchronized YamlServerConfig readFromFileOrClasspath() {
```
blootsvoets: because there is no shared data between calls (no variables are modified), the `synchronized` keyword is not necessary here. It's only necessary if a reference to the `config` variable is kept.

nivemaham: Yes. Understood. sorry. I thought I removed the synchronized here. I added it when I had the variable and forgot to remove.

PR #307, opened 2018-08-14
Move privacy url to meta token

nivemaham:
- Move privacy-policy-url to meta-token.
- This stores owning subject of the token.


blootsvoets:
https://github.com/RADAR-base/ManagementPortal/blob/d2ac91e19ffd6b156f6815d52b89742f41959d49/src/main/webapp/app/shared/subject/subject-pair-dialog.component.ts#L60 would only need to store the tokenUrl, the tokenName does not need to be in the QR code.

nivemaham:
Agreed. Would it be better that we remove it from the response itself, if we don't foresee any use of it?

PR #309, opened 2018-08-16
Added hasPermissionOnSource

### LINKED ISSUE #308: Add RadarToken.hasPermissionOnSource method
To implement RADAR-base/RADAR-Gateway#21, it would be good if `RadarToken` exposed a `hasPermissionOnSource(Permission permission, String sourceId)` method. It should allow tokens with given permission scope and:
- from a user with Permission authority in any project and with the given `sourceId` in the `sources` claim, or
- from client credentials.

blootsvoets:
Also added some javadoc and slightly modified some snippets
for readability.
Fixes #308.

### DIFF
--- a/radar-auth/src/main/java/org/radarcns/auth/token/AbstractRadarToken.java
+++ b/radar-auth/src/main/java/org/radarcns/auth/token/AbstractRadarToken.java
@@ -39,9 +39,11 @@ public abstract class AbstractRadarToken implements RadarToken {
     }
 
     @Override
-    public boolean hasPermissionOnSource(Permission permission, String sourceId) {
+    public boolean hasPermissionOnSource(Permission permission, String projectName,
+            String subjectName, String sourceId) {
         return hasScope(permission.scopeName())
-                && (isClientCredentials() || hasAuthorityForSource(permission, sourceId));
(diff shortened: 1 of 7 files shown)


radar-auth/src/main/java/org/radarcns/auth/token/AbstractRadarToken.java
```diff
@@ -25,17 +27,23 @@ public boolean hasPermission(Permission permission) {
...
     public boolean hasPermissionOnSubject(Permission permission, String projectName,
             String subjectName) {
         return hasScope(permission.scopeName())
-                && (isClientCredentials() || hasAuthorityForPermission(permission, projectName,
+                && (isClientCredentials() || hasAuthorityForSubject(permission, projectName,
                         subjectName));
     }
 
+    @Override
+    public boolean hasPermissionOnSource(Permission permission, String sourceId) {
```
nivemaham: shouldn't we validate based on projectName too,e.g. similar to `hasPermissionOnSubject`?

blootsvoets: It's just that the project and source aren't directly linked. I don't know what that would validate...

PR #312, opened 2018-08-23
Expost secret with meta token

nivemaham:
- exposes `client-secret` with Meta-token

### DIFF
--- a/src/main/java/org/radarcns/management/domain/MetaToken.java
+++ b/src/main/java/org/radarcns/management/domain/MetaToken.java
@@ -61,6 +61,9 @@ public class MetaToken extends AbstractEntity {
     @Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)
     private Subject subject;
 
+    @Column(name = "client_id", nullable = false)
+    private String clientId;
+
     /**
      * Meta token constructor.
      * Must generate a random string as the tokenName.
@@ -122,6 +125,15 @@ public class MetaToken extends AbstractEntity {
         return this;
     }
 
+    public String getClientId() {
+        return clientId;
+    }
+
+    public MetaToken clientId(String clientId) {
+        this.clientId = clientId;
+        return this;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) {
@@ -135,13 +147,15 @@ public class MetaToken extends AbstractEntity {
             && Objects.equals(tokenName, metaToken.tokenName)
             && Objects.equals(token, metaToken.token)
             && Objects.equals(fetched, metaToken.fetched)
-            && Objects.equals(expiryDate, metaToken.expiryDate);
+            && Objects.equals(expiryDate, metaToken.expiryDate)
+            && Objects.equals(clientId, metaToken.clientId)
+            && Objects.equals(subject, metaToken.subject);
     }
 
     @Override
     public int hashCode() {
 
-        return Objects.hash(id, tokenName, token, fetched, expiryDate);
+        return Objects.hash(id, tokenName, token, fetched, expiryDate, subject, clientId);
     }
 
     @Override
@@ -151,6 +165,7 @@ public class MetaToken extends AbstractEntity {
                 + ", token='" + token
                 + ", fetched=" + fetched
                 + ", expiryDate=" + expiryDate
-                + ", subject=" + subject + '}';
+                + ", subject=" + subject
+                + ", clientId=" + clientId + '}';
     }
 }
--- a/src/main/java/org/radarcns/management/service/MetaTokenService.java
+++ b/src/main/java/org/radarcns/management/service/MetaTokenService.java
@@ -1,6 +1,7 @@
 package org.radarcns.management.service;
 
 import static org.radarcns.management.web.rest.errors.EntityName.META_TOKEN;
+import static org.radarcns.management.web.rest.errors.EntityName.OAUTH_CLIENT;
 
 import java.net.MalformedURLException;
 import java.net.URL;
@@ -22,6 +23,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.security.oauth2.provider.ClientDetails;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
@@ -46,6 +48,9 @@ public class MetaTokenService {
     @Autowired
     private SubjectService subjectService;
 
+    @Autowired
+    private OAuthClientService oAuthClientService;
+
     /**
      * Save a metaToken.
      *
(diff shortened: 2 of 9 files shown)


nivemaham:
@blootsvoets @dennyverbeeck 
I realized the `client-secret` I get from ClientCredential is encoded by `BCryptPasswordEncoder`. Do you guys know how this can be decoded? 
@blootsvoets Do we prefer to transfer encoded secret over the wire, and decode it on client side?

In both cases, I could not find a way to decode this. The PasswordEncoder provides `encode` and `matches` interfaces. But I do not find anything that can decode.

blootsvoets:
If it's documented then I think it's okay. Then clients can check if they know the password, and if they don't, they can notify clients that they don't know the password. This would work well with a demo password: the client can test whether the demo password is correct, and if it is, the server works as a demo server. Otherwise, the server is off limits. Then it's up to admins to set the correct demo password.

dennyverbeeck:
Aha, yes that is true. BCrypt is a hash function, so it's one-way, you can not reverse it.

I was looking around a bit again, apparently it should be possible to configure the client secret as an empty string, which should allow that client to use the token endpoint without a client secret ([source](https://stackoverflow.com/a/48545161)). If this is the case i think this is the way to go for our public clients. What do you think @blootsvoets ?

If for some reason we can not set an empty secret, we can also configure the same secret for all public clients, e.g. `radar_client`.

dennyverbeeck:
So, i just confirmed we can use the public OAuth clients without a client secret. I left the secret field empty in the `oauth_client_details.csv` file. Then, after obtaining a metatoken and parsing the refresh token out of the endpoint response, i could use the refresh token as being the pRMT app without supplying a client secret, as seen by the below `curl` call and response:

```
curl -XPOST http://pRMT@localhost:8080/oauth/token -F "grant_type=refresh_token" -F "refresh_token=$rtoken"
{
  "access_token" : "eyJ0eXAiOiJKV1QiLCJhbGciOiJFUzI1NiJ9.eyJzdWIiOiJzdWItMSIsInNvdXJjZXMiOlsiMDNkMjhlNWMtZTAwNS00NmQ0LWE5YjMtMjc5YzI3ZmJiYzgzIl0sInVzZXJfbmFtZSI6InN1Yi0xIiwicm9sZXMiOlsicmFkYXI6Uk9MRV9QQVJUSUNJUEFOVCJdLCJpc3MiOiJNYW5hZ2VtZW50UG9ydGFsIiwiYXV0aG9yaXRpZXMiOlsiUk9MRV9QQVJUSUNJUEFOVCJdLCJjbGllbnRfaWQiOiJwUk1UIiwiYXVkIjpbInJlc19nYXRld2F5IiwicmVzX01hbmFnZW1lbnRQb3J0YWwiXSwiZ3JhbnRfdHlwZSI6ImF1dGhvcml6YXRpb25fY29kZSIsInNjb3BlIjpbIk1FQVNVUkVNRU5ULkNSRUFURSIsIlNVQkpFQ1QuVVBEQVRFIiwiU1VCSkVDVC5SRUFEIiwiUFJPSkVDVC5SRUFEIiwiU09VUkNFVFlQRS5SRUFEIiwiU09VUkNFLlJFQUQiLCJTT1VSQ0VEQVRBLlJFQUQiLCJVU0VSLlJFQUQiLCJST0xFLlJFQUQiXSwiZXhwIjoxNTM1MDg2OTg3LCJpYXQiOjE1MzUwNDM3ODcsImp0aSI6IjdjNTI3MDQ4LTM3NzUtNDk1MS1iN2IzLTliNjhkMTZhMTJjOSJ9.MEUCIFKeuttWxs-TXDAyImKSWR2K5mIHRuW9jmAXrZaZtFA2AiEA4W5MCCoOIGJVXrK1k5rnEHwS_XJasnSCDj4Wz0jyEO4",
  "token_type" : "bearer",
  "refresh_token" : "eyJ0eXAiOiJKV1QiLCJhbGciOiJFUzI1NiJ9.eyJzdWIiOiJzdWItMSIsInNvdXJjZXMiOlsiMDNkMjhlNWMtZTAwNS00NmQ0LWE5YjMtMjc5YzI3ZmJiYzgzIl0sInVzZXJfbmFtZSI6InN1Yi0xIiwicm9sZXMiOlsicmFkYXI6Uk9MRV9QQVJUSUNJUEFOVCJdLCJpc3MiOiJNYW5hZ2VtZW50UG9ydGFsIiwiYXV0aG9yaXRpZXMiOlsiUk9MRV9QQVJUSUNJUEFOVCJdLCJjbGllbnRfaWQiOiJwUk1UIiwiYXVkIjpbInJlc19nYXRld2F5IiwicmVzX01hbmFnZW1lbnRQb3J0YWwiXSwiZ3JhbnRfdHlwZSI6ImF1dGhvcml6YXRpb25fY29kZSIsInNjb3BlIjpbIk1FQVNVUkVNRU5ULkNSRUFURSIsIlNVQkpFQ1QuVVBEQVRFIiwiU1VCSkVDVC5SRUFEIiwiUFJPSkVDVC5SRUFEIiwiU09VUkNFVFlQRS5SRUFEIiwiU09VUkNFLlJFQUQiLCJTT1VSQ0VEQVRBLlJFQUQiLCJVU0VSLlJFQUQiLCJST0xFLlJFQUQiXSwiYXRpIjoiN2M1MjcwNDgtMzc3NS00OTUxLWI3YjMtOWI2OGQxNmExMmM5IiwiZXhwIjoxNTQyOTkyNTg3LCJpYXQiOjE1MzUwNDM3ODcsImp0aSI6ImI5Yjc5YWRhLWIzMDMtNDk1MC05OWE2LTUzYmU1NDkyNjRmNSJ9.MEYCIQCT34GEDz77Z1CUVFRJXU4xBmJwNQYcXRCGW0xdWaoFDQIhAJwzFUMjUeKST5ZwCCiY2Rr0UlMmCnqgs0ZUV_c5tBSc",
  "expires_in" : 43199,
  "scope" : "MEASUREMENT.CREATE SUBJECT.UPDATE SUBJECT.READ PROJECT.READ SOURCETYPE.READ SOURCE.READ SOURCEDATA.READ USER.READ ROLE.READ",
  "sub" : "sub-1",
  "sources" : [ "03d28e5c-e005-46d4-a9b3-279c27fbbc83" ],
  "grant_type" : "authorization_code",
  "roles" : [ "radar:ROLE_PARTICIPANT" ],
  "iss" : "ManagementPortal",
  "iat" : 1535043787,
  "jti" : "7c527048-3775-4951-b7b3-9b68d16a12c9"
}
```

Therefore i think we should go with this approach, as it adheres to the OAuth spec recommendation of not using a client secret at all with public clients.

I'll open a new PR to take out again the client secret from the `TokenDTO` class. I think the other changes in this PR are fine. Storing the client id with the metatoken can be useful as an extra security measure, if you need to know the token URL AND the client id this further increases the difficulty to brute-force the metatoken endpoint.

blootsvoets:
@dennyverbeeck sounds fine. It requires a few client changes:
- the defaults in this repo and in RADAR-docker should have empty secrets for clients with `authorized_code` and without `client_credentials` or `password`.
- the defaults for the RADAR clients should also have an empty password.

PR #318, opened 2018-08-29
Remove client secret field from metatoken

dennyverbeeck:
- No longer include client secret in the `/api/meta-token` response
- Make client secrets empty for public OAuth clients
- Update documentation
- Add an `ExceptionTranslator` for Spring Security OAuth's `NoSuchClientException`: trying to initiate an OAuth flow with a non-existing client would result in an `Internal Server Error` response, now it is translated to a `Bad Request` response


README.md
```diff
@@ -121,37 +121,48 @@ clients for an example.
...
+     where you replace `MyId` with your OAuth client id. This needs to be done from a interactive 
+     web view, either a browser or a web window. If the user approves, this will redirect to 
+     `https://my.example.com/oauth_redirect?code=abcdef`. In Android, with [https://appauth.io]
+     (AppAuth library), the URL could be `com.example.my://oauth_redirect` for the `com.example.my`
+      app.
+2. Request a token for your app by doing a POST, again with HTTP basic authentication with as 
+username your OAuth client id, and leaving the password empty:
     ```
-    POST MyId:MySecret /oauth/token
+    POST /oauth/token
```
blootsvoets: Doesn't this need to include the client ID as a username, e.g. `MyId@/oauth/token`?

PR #326, opened 2018-09-20
Update frontend dep

nivemaham:
upgrading angular version to 4.1.3


blootsvoets:
Could you add the `yarn.lock` and remove the `package.lock`? Run `yarn install` to generate it.

nivemaham:
Yep. Yarn.lock is already added to git. I rolled back the package-lock.json

PR #342, opened 2018-11-28
improve activation email with configurable timeout 

nivemaham:
- improve activation email with configurable timeout for activation/reset key 
- fix message for reset password request

### DIFF
--- a/src/main/java/org/radarcns/management/service/MailService.java
+++ b/src/main/java/org/radarcns/management/service/MailService.java
@@ -104,7 +104,7 @@ public class MailService {
         context.setVariable(USER, user);
         context.setVariable(BASE_URL,
                 managementPortalProperties.getCommon().getManagementPortalBaseUrl());
-        context.setVariable(EXPIRY, Duration.ofSeconds(duration).toHours() + " Hours");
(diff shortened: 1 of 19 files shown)


src/main/java/org/radarcns/management/service/MailService.java
```diff
@@ -94,13 +97,14 @@ public void sendActivationEmail(User user) {
...
     @Async
-    public void sendCreationEmail(User user) {
+    public void sendCreationEmail(User user, long duration) {
         log.debug("Sending creation email to '{}'", user.getEmail());
         Locale locale = Locale.forLanguageTag(user.getLangKey());
         Context context = new Context(locale);
         context.setVariable(USER, user);
         context.setVariable(BASE_URL,
                 managementPortalProperties.getCommon().getManagementPortalBaseUrl());
+        context.setVariable(EXPIRY, Duration.ofSeconds(duration).toHours() + " Hours");
```
blootsvoets: leave the Hours out, use lowercase, and move it to the language-specific templates (use "uur" in Dutch). Alternatively, use a language mapping in this method, if that is easy to do.

PR #347, opened 2018-12-05
show meta-token timeout info to the user.

### LINKED ISSUE #346: Meta QR code expiry
The low time period for the expiry of the meta token (1hr by default) means a lot of remote recruitments having problems.

It would be good if the MP displayed the expiry/timeout of the Token URL on the pair app dialog so the clinicians can make informed decisions.

Would also be good to have this configurable by admins on the MP (don't know if its already there but couldn't find any docs)

nivemaham:
Fixes #346

### DIFF
--- a/src/main/webapp/i18n/en/subject.json
+++ b/src/main/webapp/i18n/en/subject.json
@@ -66,7 +66,7 @@
             "hideTokenUrl" : "Hide token data",
             "baseUrl" : "Base Url",
             "tokenName" : "Token Name",
-            "tokenTimeoutMessage": "Please note that this QR code content will expire in {{time}} hour(s). <br>Please refresh to renew the QR code."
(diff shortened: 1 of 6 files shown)


src/main/webapp/i18n/en/subject.json
```diff
@@ -65,7 +65,8 @@
             "showTokenUrl" : "Show token data",
             "hideTokenUrl" : "Hide token data",
             "baseUrl" : "Base Url",
-            "tokenName" : "Token Name"
+            "tokenName" : "Token Name",
+            "tokenTimeoutMessage": "Please note that this QR code content will expire in {{time}} hour(s). <br>Please refresh to renew the QR code."
```
blootsvoets: Instead (or in addition) to the time-out, it would be good to show the time of the user, when it will time-out. For US that would be Dec 5, 20:17 PM, for NL 5 dec, 20:17 uur.

PR #396, opened 2019-02-20
Upgrade Mapstruct support and other dependencies 

nivemaham:
- Upgraded Mapstruct to 1.3.0 and using annotationProcessor from gradle
- Modified `POST /api/subjects` body to contain just the `projectName`, instead of complete `ProjectDTO`. 
- modified test cases to use unique project-id, external-id combination.
- reduce test-case complexity.
- Upgrade front-end and back-dependencies based on `dependabot` suggestions.


nivemaham:
@blootsvoets Sorry. I accidentally closed the previous PR.

nivemaham:
Closing this PR based on offline discussions. The version upgrades except `mapstruct` are merged with #403 . The API change on Subject GET interface will break the existing integration and this change is not necessary as of now.

PR #403, opened 2019-02-27
Upgrade auth0 version

nivemaham:
- Upgrade auth0 version and uses `JWTVerifier` instead of `SignatureVerifier`
- Use custom `JwtAccessTokenConverter` and `TokenStore` (Default TokenStore is based on Spring JwtAccessTokenConverter class)
- Refactor KeyStore related functionalities, 
- Use a single point to handle all Keystore related functions. 
- Reduce code complexity and code redundancy


nivemaham:
Thanks for the comments. I agree that the API change can break existing integration. Both changes are not urgent. I will split them into two PRs.  There is a separate PR #396  which has just the dependency upgrades and API change due to mapstruct upgrade. 
The API change was necessary due to the mapstruct version upgrade. Current mappers expects all of the nested children to exist, which makes it very complex to test. I also thought it would be efficient to use just the projectName and fetch the project details using a separate request. I agree on making it a major release. Let's continue the discussion on other PR thread.

PR #406, opened 2019-03-01
fix multiple roles shown for user

### LINKED ISSUE #388: Project Admins displayed multiple times
Some UI issue in MP i think. Its showing each admin multiple times for a project.

screenshot 2019-02-14 at 15 53 02
screenshot 2019-02-14 at 15 55 05

nivemaham:
Current User Criteria builder does not return unique results with multiple parameter queries. Fixes it by adding distinct-> true 
Fixes #388

### DIFF
--- a/src/main/java/org/radarcns/management/repository/filters/UserFilter.java
+++ b/src/main/java/org/radarcns/management/repository/filters/UserFilter.java
@@ -52,7 +52,7 @@ public class UserFilter implements Specification<User> {
             return null;
         } else {
             query.distinct(true);
-            return cb.and(predicates.toArray(new Predicate[predicates.size()]));
+            return cb.and(predicates.toArray(new Predicate[0]));
         }
     }
 
--- a/src/main/java/org/radarcns/management/repository/filters/UserFilter.java
+++ b/src/main/java/org/radarcns/management/repository/filters/UserFilter.java
@@ -29,12 +29,10 @@ public class UserFilter implements Specification<User> {
         if (StringUtils.isNotBlank(login)) {
             predicates
                 .add(cb.like(cb.lower(root.get("login")), "%" + login.trim().toLowerCase() + "%"));
-            query.distinct(true);
         }
         if (StringUtils.isNotBlank(email)) {
             predicates
                 .add(cb.like(cb.lower(root.get("email")), "%" + email.trim().toLowerCase() + "%"));
-            query.distinct(true);
         }
 
         if (StringUtils.isNotBlank(projectName)) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/radarcns/management/repository/filters/UserFilter.java
```diff
@@ -29,23 +29,27 @@ public Predicate toPredicate(Root<User> root, CriteriaQuery<?> query, CriteriaBu
...
+            query.distinct(true);
         }
         if (StringUtils.isNotBlank(authority)) {
             Join<User, Role> roleJoin = root.join("roles");
             Join<Role, Authority> authorityJoin = roleJoin.join("authority");
             predicates.add(cb.like(cb.lower(authorityJoin.get("name")),
                     "%" + authority.trim().toLowerCase() + "%"));
+            query.distinct(true);
         }
         return predicates.isEmpty() ? null
```
blootsvoets: Instead of the four statements, perhaps make this an if-else?
```java
if (predicates.isEmpty()) {
    return null
} else {
    query.distinct(true);
    return cb.and(predicates.toArray(new Predicate[0]));
}
```

nivemaham: Makes sense. Thanks :)

src/main/java/org/radarcns/management/repository/filters/UserFilter.java
```diff
@@ -47,8 +47,13 @@ public Predicate toPredicate(Root<User> root, CriteriaQuery<?> query, CriteriaBu
...
                     "%" + authority.trim().toLowerCase() + "%"));
         }
-        return predicates.isEmpty() ? null
-                : cb.and(predicates.toArray(new Predicate[predicates.size()]));
+
+        if (predicates.isEmpty()) {
+            return null;
+        } else {
+            query.distinct(true);
+            return cb.and(predicates.toArray(new Predicate[predicates.size()]));
```
blootsvoets: For future reference: `predicates.toArray(new Predicate[0])` is faster to execute and shorter than `predicates.toArray(new Predicate[predicates.size()])`

nivemaham: Apparently PMD 1.8 advises other way around. 
Calls to a collection's toArray() method should specify target arrays sized to match the size of the collection. Initial arrays that are too small are discarded in favour of new ones that have to be created that are the proper size.
```
List foos = getFoos();

    // inefficient, the array will be discarded
Foo[] fooArray = foos.toArray(new Foo[0]);

    // much better; this one sizes the destination array,
    // avoiding of a new one via reflection
Foo[] fooArray = foos.toArray(new Foo[foos.size()]);
```

blootsvoets: It used to be like that when PMD 1.8 came out, but since Java 5 the initialiser with 0 is faster. For a full review: https://shipilev.net/blog/2016/arrays-wisdom-ancients/ ([Conclusion](https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_conclusion))

nivemaham: Yes. I read about it. I will also look into upgrading PMD version

blootsvoets: Probably the rule is still in there, you can disable it in the performance ruleset, by adding
```xml
<exclude rule="OptimizableToArrayCall"/>
```

PR #409, opened 2019-03-04
Fix subject-source assignment getting lost after a source is edited.

### LINKED ISSUE #345: Subject-source relationship disappears after a source is edited.
Source-A is assigned to Subject A. If I edit Source-A, Source-A is not listed as assigned sources under Subject A in project view.

nivemaham:
The old `ManyToMany` mapping subject-source relationship was maintained in a separate table with two foreign Keys. After moving to `OneToMany` source -> subject relationship is maintained by a foreign-key in source. However, the update method does not consider this change. 
During update of a source, assigned subject gets lost due to the existing implementation of the mapper where subject is ignored. 
Fix is made by explicitly setting the subject from existing source or by subjectLogin. 

Note: If an already assigned source was edited for some reason, the corresponding subject to whom the source was assigned is lost. This would have created inconsistant data which says a source is assigned, but cannot find the assigned subject.  
Fixes #345.

### DIFF
--- a/src/main/java/org/radarcns/management/domain/Subject.java
+++ b/src/main/java/org/radarcns/management/domain/Subject.java
@@ -65,7 +65,7 @@ public class Subject extends AbstractEntity implements Serializable {
     @Cascade(CascadeType.ALL)
     private User user;
 
-    @OneToMany(mappedBy = "subject")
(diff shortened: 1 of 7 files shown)


blootsvoets:
LGTM

PR #415, opened 2019-03-12
Fix jwt issues

nivemaham:
Adding hot-fixes for current release branch to make the future release stable.

### DIFF
--- a/radar-auth/src/test/java/org/radarcns/auth/util/TokenTestUtils.java
+++ b/radar-auth/src/test/java/org/radarcns/auth/util/TokenTestUtils.java
@@ -1,8 +1,5 @@
 package org.radarcns.auth.util;
 
-import com.auth0.jwt.JWT;
-import com.auth0.jwt.algorithms.Algorithm;
-import com.auth0.jwt.interfaces.DecodedJWT;
 import java.io.InputStream;
 import java.security.KeyStore;
 import java.security.cert.Certificate;
@@ -11,6 +8,10 @@ import java.security.interfaces.RSAPublicKey;
 import java.time.Instant;
 import java.util.Date;
 import java.util.stream.Collectors;
+
+import com.auth0.jwt.JWT;
+import com.auth0.jwt.algorithms.Algorithm;
+import com.auth0.jwt.interfaces.DecodedJWT;
 import org.apache.commons.codec.binary.Base64;
 import org.radarcns.auth.authorization.Permission;
 
@@ -63,9 +64,11 @@ public class TokenTestUtils {
     }
 
     private static void initVars(String publicKey, Algorithm algorithm) {
-        PUBLIC_KEY_BODY = "[\n"
-            + " \"-----BEGIN PUBLIC KEY-----\\n" + publicKey + "\\n-----END PUBLIC KEY-----\"\n"
-            + "]";
+        PUBLIC_KEY_BODY =
+                "{\n" + " \"keys\" : [ {\n" + "  \"alg\" : \"" + algorithm.getName() + "\",\n"
+                        + "  \"alg\" : \"" + "RSA" + "\",\n"
+                        + "  \"value\" : \"-----BEGIN PUBLIC KEY-----\\n" + publicKey
+                        + "\\n-----END PUBLIC " + "KEY-----\"\n" + "} ]\n" + "}";
 
         Instant exp = Instant.now().plusSeconds(30 * 60);
         Instant iat = Instant.now();
--- a/radar-auth/src/test/java/org/radarcns/auth/util/TokenTestUtils.java
+++ b/radar-auth/src/test/java/org/radarcns/auth/util/TokenTestUtils.java
@@ -63,11 +63,9 @@ public class TokenTestUtils {
     }
 
     private static void initVars(String publicKey, Algorithm algorithm) {
-        PUBLIC_KEY_BODY = "{\n"
-            + "  \"alg\" : \"SHA256withRSA\",\n"
-            + "  \"value\" : \"-----BEGIN PUBLIC KEY-----\\n" + publicKey + "\\n-----END PUBLIC "
-            + "KEY-----\"\n"
-            + "}";
+        PUBLIC_KEY_BODY = "[\n"
+            + " \"-----BEGIN PUBLIC KEY-----\\n" + publicKey + "\\n-----END PUBLIC KEY-----\"\n"
+            + "]";
 
         Instant exp = Instant.now().plusSeconds(30 * 60);
         Instant iat = Instant.now();
--- a/src/main/java/org/radarcns/management/config/OAuth2ServerConfiguration.java
+++ b/src/main/java/org/radarcns/management/config/OAuth2ServerConfiguration.java
@@ -241,7 +241,7 @@ public class OAuth2ServerConfiguration {
         public void configure(AuthorizationServerSecurityConfigurer oauthServer) {
             oauthServer.allowFormAuthenticationForClients()
                     .checkTokenAccess("isAuthenticated()")
-                    .tokenKeyAccess("isAnonymous() || isAuthenticated()")
+                    .tokenKeyAccess("permitAll()")
                     .passwordEncoder(new BCryptPasswordEncoder());
         }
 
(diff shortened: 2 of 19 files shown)


src/main/java/org/radarcns/management/security/jwt/ManagementPortalJwtAccessTokenConverter.java
```diff
@@ -150,24 +150,15 @@ public OAuth2AccessToken enhance(OAuth2AccessToken accessToken,
...
             // Refresh tokens do not expire unless explicitly of the right type
             refreshTokenToEnhance.setExpiration(null);
-
-            // I don't really get what is the point of doing this.
-            try {
-                Map<String, Object> claims = objectMapper
-                        .parseMap(JwtHelper.decode(refreshToken.getValue()).getClaims());
-                if (claims.containsKey(TOKEN_ID)) {
-                    refreshTokenToEnhance.setValue(claims.get(TOKEN_ID).toString());
-                }
```
blootsvoets: If I understand correctly, isn't this the code to allow a `TOKEN_ID` to be invalidated in the `approvalStore()`? The refresh token itself will change every update but this would allow a certain refresh token ID to be invalidated, both for the access token and the refresh token.

src/main/java/org/radarcns/management/config/OAuth2ServerConfiguration.java
```diff
@@ -201,7 +201,7 @@ public TokenEnhancer tokenEnhancer() {
 
         @Bean
         public TokenStore tokenStore() {
-            return new ManagementPortalJwtTokenStore(accessTokenConverter(), approvalStore());
```
blootsvoets: Would this `approvalStore()` not allow the MP store to invalidate refresh tokens?

src/main/java/org/radarcns/management/web/rest/TokenKeyEndpoint.java
```diff
@@ -0,0 +1,53 @@
...
+     */
+    @GetMapping("/oauth/token_key")
+    @Timed
+    public List<String> getKey(Principal principal) throws NotAuthorizedException {
+        logger.debug("Requesting verifier public keys...");
+        if ((principal == null || principal instanceof AnonymousAuthenticationToken)) {
+            throw new NotAuthorizedException("You need to authenticate to see a shared key");
+        }
+
+        return managementPortalOauthKeyStoreHandler.loadVerifyingPublicKeys();
```
blootsvoets: If we're going to expose multiple web keys, why not use the JWKS spec: https://auth0.com/docs/jwks
Detailed JSON: https://auth0.com/docs/tokens/reference/jwt/jwks-properties

blootsvoets: Example from the spec:
```JSON
{"keys":
       [
         {"kty":"EC",
          "crv":"P-256",
          "x":"MKBCTNIcKUSDii11ySs3526iDZ8AiTo7Tu6KPAqv7D4",
          "y":"4Etl6SRW2YiLUrN5vfvVHuhp7x8PxltmWWlbbM4IFyM",
          "use":"enc",
          "kid":"1"},

         {"kty":"RSA",
          "n": "0vx7agoebGcQSuuPiLJXZptN9nndrQmbXEps2aiAFbWhM78LhWx
     4cbbfAAtVT86zwu1RK7aPFFxuhDR1L6tSoc_BJECPebWKRXjBZCiFV4n3oknjhMs
     tn64tZ_2W-5JsGY4Hc5n9yBXArwl93lqt7_RN5w6Cf0h4QyQ5v-65YGjQR0_FDW2
     QvzqY368QQMicAtaSqzs8KJZgnYb9c7d0zgdAZHzu6qMQvRL5hajrn1n91CbOpbI
     SD08qNLyrdkt-bFTWhAI4vMQFh6WeZu0fM4lFd2NcRwr3XPksINHaQ-G_xBniIqb
     w0Ls1jF44-csFCur-kEgU8awapJzKnqDKgw",
          "e":"AQAB",
          "alg":"RS256",
          "kid":"2011-04-29"}
       ]
     }
```

src/main/java/org/radarcns/management/config/OAuth2ServerConfiguration.java
nivemaham: I was trying to stick to exactly what the Spring Implementation was doing. There the JwtTokenStore is not actually a "Store" per say, they do not use it to store or invalidate a token using that. The actual approval store in action is the JdbcApprovalStore or the PostGreApprovalStore. However, it is not used in JwtTokenStore based on the existing implementation.

src/main/java/org/radarcns/management/security/jwt/ManagementPortalJwtAccessTokenConverter.java
nivemaham: Technically yes.It is the id. I am also little confused about this implementation. Main confusion is sometimes the value is set to a UUID which is the token-id and sometimes it is the encoded string. Based on some debugging it appears that the value we get at this point `JwtHelper.decode(refreshToken.getValue()).getClaims()` is a UUID and we cannot decode it as a JWT. so It always fails here. I will revert the changes to fit exactly what was there earlier. I will look into this again.

src/main/java/org/radarcns/management/web/rest/TokenKeyEndpoint.java
nivemaham: So I had a look into JWKS spec and possibilities to implement it on MP. 
From what I read every public-key needs a unique-id as an identifier and it is added as 'kid' in the header. I do not think it is the case right now. But probably there is a way to do that easily. However, the bigger confusion is current implementation just needs the public-key to create a the algorithm the the verifier. So we have to change the implementation to use JWKProvider. 
Plus oauth0, has jwk library for [rsa](https://github.com/auth0/jwks-rsa-java) only. I think we need both rsa and EC.
Other option is to use [Nimbus JOSE+JWT library ](https://connect2id.com/products/nimbus-jose-jwt/examples/jwk-retrieval), assuming we will replace the use of oauth0 with Nimbus. What do you think? @blootsvoets 
Am I missing something?

blootsvoets: One implementation of the `kid` is to use the SHA-1 sum of the key. This way no additional bookkeeping is needed. If auth0 doesn't support it, eventually probably Nimbus would be a better choice. For now, we can stick to auth0, but perhaps do match the general layout:
```json
{"keys":
       [
         {
          "alg": <algorithm>,
          "value": <Auth0 encoded public key>
         }
       ]
}
```

nivemaham: Make sense. I agree. For now I will change the API to fit what is proposed here. Later, we can investigate on migrating to Nimbus. I have one last question. Just to confirm.
Should the `oauth/token_key` an openly available endpoint i.e. no authentication? I think it should not be. Currently we expect some valid Principal to be available. 
Assuming only registered clients will request for JWKs, can I use basic auth with the client-credentials to request JWK? @blootsvoets

blootsvoets: In OpenID connect, public keys are considered public, so no authentication should be necessary. This is generally true, public keys can always be exposed.

radar-auth/src/test/java/org/radarcns/auth/util/TokenTestUtils.java
```diff
@@ -63,11 +63,9 @@ public static void setUp() throws Exception {
...
 
     private static void initVars(String publicKey, Algorithm algorithm) {
-        PUBLIC_KEY_BODY = "{\n"
-            + "  \"alg\" : \"SHA256withRSA\",\n"
-            + "  \"value\" : \"-----BEGIN PUBLIC KEY-----\\n" + publicKey + "\\n-----END PUBLIC "
-            + "KEY-----\"\n"
-            + "}";
+        PUBLIC_KEY_BODY = "[\n"
+            + " \"-----BEGIN PUBLIC KEY-----\\n" + publicKey + "\\n-----END PUBLIC KEY-----\"\n"
+            + "]";
```
blootsvoets: Perhaps change this test as well? For the `alg` parameter, the spec uses values from [RFC 7518](https://www.rfc-editor.org/rfc/rfc7518.txt), which translates to `alg = RS256` and `kty = RSA` for RSA 256, and `alg = ES256` and `kty = EC` for ECDSA with `P-256` curve.

nivemaham: Yes. Will do

PR #421, opened 2019-03-19
remove annotationProcessor in compile 

nivemaham:
remove apt plugin dependency and use annotationProcessor instead of compile dependency

### DIFF
--- a/gradle/mapstruct.gradle
+++ b/gradle/mapstruct.gradle
@@ -1,5 +1,6 @@
-apply plugin: "net.ltgt.apt"
 
 dependencies {
-    apt "org.mapstruct:mapstruct-processor:${mapstruct_version}"
+    compile group: 'org.mapstruct', name: 'mapstruct-jdk8', version: mapstruct_version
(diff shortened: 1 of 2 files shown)


gradle/mapstruct.gradle
```diff
@@ -1,5 +1,6 @@
-apply plugin: "net.ltgt.apt"
 
 dependencies {
-    apt "org.mapstruct:mapstruct-processor:${mapstruct_version}"
+    compile group: 'org.mapstruct', name: 'mapstruct-jdk8', version: mapstruct_version
+    compileOnly group: 'org.mapstruct', name: 'mapstruct-processor', version: mapstruct_version
+    annotationProcessor group: 'org.mapstruct', name: 'mapstruct-processor', version: mapstruct_version
```
blootsvoets: Are both `compileOnly` and `annotationProcessor` needed?

nivemaham: Good point.Forgot to remove it.

PR #425, opened 2019-03-21
Auth0 validation with shading

nivemaham:
- Adding deprecated `java-jwt:3.2.0` dependency to keep validating old tokens. 
- Deprecate configuring public-keys and use public-key endpoint.

**Changes for migration from 0.5.3 to 0.5.5**
When upgrading from versions earlier than 0.5.4, please follow these steps.
1. Copy old public-key file to MP class-path and configure environment variable `RADAR_IS_CONFIG_LOCATION` to the location of the config file provided.
2. Set `MANAGEMENTPORTAL_OAUTH_ENABLE_PUBLIC_KEY_VERIFIERS` to `"true"`.
3. If you have other components who validates the token, use the public-key endpoint to request for public-keys in the form of `JavaWebKeySet`. Please upgrade your `radar-auth` dependency to at lease `0.5.4` or higher.

### DIFF
--- a/radar-auth/src/main/java/org/radarcns/auth/authentication/AlgorithmLoader.java
+++ b/radar-auth/src/main/java/org/radarcns/auth/authentication/AlgorithmLoader.java
@@ -18,21 +18,21 @@ import org.radarcns.auth.token.validation.deprecated.DeprecatedEcTokenValidation
 public class AlgorithmLoader {
 
     private final List<TokenValidationAlgorithm> supportedAlgorithmsForWebKeySets;
-    private final List<TokenValidationAlgorithm> supportedAlgorithmsForPublicKeys;
+    private final List<TokenValidationAlgorithm> supportedAlgorithmsForPublicKeysInConfig;
 
     /**
      * Creates an instance of {@link AlgorithmLoader} with lists of
      * {@link TokenValidationAlgorithm} provided.
-     * @param algorithmsForWebKeys default support. Algorithms to be supported for
+     * @param supportedAlgorithmsForWebKeySets default support. Algorithms to be supported for
      *                             public keys shared from public key endpoints as
      *                             {@link JavaWebKeySet}.
-     * @param algorithmsForLocalPublicKeys deprecated support. Algorithms to be supported for
-     *                                     public keys configured in config file.
+     * @param supportedAlgorithmsForPublicKeysInConfig deprecated support. Algorithms to be
+     *                                      supported for public keys configured in config file.
      */
-    public AlgorithmLoader(List<TokenValidationAlgorithm> algorithmsForWebKeys,
-            List<TokenValidationAlgorithm> algorithmsForLocalPublicKeys) {
-        this.supportedAlgorithmsForWebKeySets = algorithmsForWebKeys;
-        this.supportedAlgorithmsForPublicKeys = algorithmsForLocalPublicKeys;
+    public AlgorithmLoader(List<TokenValidationAlgorithm> supportedAlgorithmsForWebKeySets,
+            List<TokenValidationAlgorithm> supportedAlgorithmsForPublicKeysInConfig) {
+        this.supportedAlgorithmsForWebKeySets = supportedAlgorithmsForWebKeySets;
+        this.supportedAlgorithmsForPublicKeysInConfig = supportedAlgorithmsForPublicKeysInConfig;
     }
 
     /**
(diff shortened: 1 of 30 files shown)


radar-auth/src/main/java/org/radarcns/auth/authentication/AlgorithmLoader.java
```diff
@@ -0,0 +1,81 @@
...
+import com.auth0.jwt.algorithms.Algorithm;
+import org.radarcns.auth.exception.TokenValidationException;
+import org.radarcns.auth.security.jwk.JavaWebKey;
+import org.radarcns.auth.security.jwk.JavaWebKeySet;
+import org.radarcns.auth.token.validation.ECTokenValidationAlgorithm;
+import org.radarcns.auth.token.validation.RSATokenValidationAlgorithm;
+import org.radarcns.auth.token.validation.TokenValidationAlgorithm;
+import org.radarcns.auth.token.validation.deprecated.DeprecatedEcTokenValidationAlgorithm;
+
+public class AlgorithmLoader {
```
blootsvoets: This could have a constructor, with the fields and methods not static. That would allow it to be loaded via Spring if needed.

radar-auth/src/main/java/org/radarcns/auth/authentication/AlgorithmLoader.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private static final List<TokenValidationAlgorithm> algorithmList = Arrays.asList(
+            new ECTokenValidationAlgorithm(),
+            new RSATokenValidationAlgorithm());
+    private static final List<TokenValidationAlgorithm> supportedAlgorithmsForPublicKeys =
+            Arrays.asList(
```
blootsvoets: By making this a method, it can be overridden.

radar-auth/src/main/java/org/radarcns/auth/authentication/AlgorithmLoader.java
```diff
@@ -0,0 +1,81 @@
...
+            new DeprecatedEcTokenValidationAlgorithm(),
```
blootsvoets: Both the deprecated and new EcToken validation should be supported for public keys.

radar-auth/src/main/java/org/radarcns/auth/authentication/AlgorithmLoader.java
nivemaham: Good point. I will change it. Loading is done in constructor when the spring component starts .

radar-auth/src/main/java/org/radarcns/auth/authentication/AlgorithmLoader.java
nivemaham: Sorry. I am not sure whether the variable names were misleading. RSA and new EC are supported for JWKs and RSA and DeprecatedEC are supported for loading public-keys configured in files, which i consider the "deprecated way". Did you mean that new EC and Deprecated EC should be supported for loading public-keys from config-file?
Having both EC versions in the public-key support may lead to some confusions, since we do not know how to decide whether a public-key is for new EC or Deprecated EC. We can only decide whether it is EC key or not. We can indeed try till there is a successful loading. But I think we also don't have a need for that.  
I made the implementation based on the concept of, if there is a EC public key provided in config file, it is generated from Deprecated EC method and all the new tokens/implementations should depend on public key endpoint to fetch public-keys. This way we can be sure that after a single iteration of refreshing old tokens all of the tokens have migrated to the new EC method and the public-key endpoint can be used for validation.

nivemaham:
Thanks. I have updated the PR with details on how to migrate from older version. 
If we want to rotate the key for signing, create a new key with a unique alias on the keystore.p12 used by MP. 
Add the alias to `checkingKeyAliases` config.
 e.g.
`MANAGEMENTPORTAL_OAUTH_CHECKING_KEY_ALIASES_2: new-alias`
on the `docker-compose.yml`
` SPRING_APPLICATION_JSON: '{"managementportal":{"oauth":{"checkingKeyAliases":["${MANAGEMENTPORTAL_OAUTH_CHECKING_KEY_ALIASES_0}","${MANAGEMENTPORTAL_OAUTH_CHECKING_KEY_ALIASES_1}", "${MANAGEMENTPORTAL_OAUTH_CHECKING_KEY_ALIASES_2}"]}}}'`

PR #426, opened 2019-03-25
Release 0.5.5

nivemaham:
**Changes from release 0.5.4**
- update some dependencies
- Upgrade to gradle `5.2.1`
- Add support for loading old public-keys for verification from config files and validate using `java-jwt:3.2.0`.

**Migrating from 0.5.3 to 0.5.5**
When upgrading from versions earlier than 0.5.4, please follow these steps.
1. Copy old public-key file to MP class-path and configure environment variable `RADAR_IS_CONFIG_LOCATION` to the location of the config file provided.
2. Set `MANAGEMENTPORTAL_OAUTH_ENABLE_PUBLIC_KEY_VERIFIERS` to `"true"`.
3. If you have other components who validates the token, use the public-key endpoint to request for public-keys in the form of `JavaWebKeySet`. Please upgrade your `radar-auth` dependency to at lease `0.5.4` or higher.


radar-auth/settings.gradle
```diff
@@ -1 +1,2 @@
 rootProject.name = 'radar-auth'
+include 'deprecated-auth0'
```
blootsvoets: This file should probably be removed. Only the root needs a `settings.gradle`.

PR #428, opened 2019-04-09
Fix bintray upload issues

nivemaham:
- Add publishing to shadowed jar
- Add publishing directly on sub-projects to avoid creating invalid pom files


nivemaham:
That was indeed my expectation too. However from what i have checked, the radar-auth-*-all.jar does not contain the shadowed classes. I decided to publish the deprecated shadowed library, just in case it becomes necessary.
 - [ ] radar-auth jar contains shadowed auth0 and all its dependencies 
 - [x] radar-auth pom does not reference these dependencies
 - [x] radar-auth javadoc and sources also do not reference these dependencies

Issue #434, opened 2019-05-21
Reconnect on Postgresql error


k1-hedayati:
Management portal should not crash if Postgresql connection cannot be made, it should try to reconnect.

```
2019-05-21 07:11:13.465  WARN 1 --- [  XNIO-2 task-3] o.h.engine.jdbc.spi.SqlExceptionHelper   : SQL Error: 0, SQLState: null                                                                                      
2019-05-21 07:11:13.465 ERROR 1 --- [  XNIO-2 task-3] o.h.engine.jdbc.spi.SqlExceptionHelper   : Connection is closed                                                                                              
2019-05-21 07:11:13.467 ERROR 1 --- [  XNIO-2 task-3] o.r.m.w.rest.errors.ExceptionTranslator  : Failed to process message                                                                                         
                                                                                                                                                                                                                   
javax.persistence.PersistenceException: org.hibernate.exception.GenericJDBCException: could not prepare statement          
        at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:149)                  
        at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:157)                     
        at org.hibernate.query.internal.AbstractProducedQuery.list(AbstractProducedQuery.java:1423)                
        at org.hibernate.envers.query.internal.impl.AbstractAuditQuery.buildAndExecuteQuery(AbstractAuditQuery.java:101)                 
        at org.hibernate.envers.query.internal.impl.RevisionsOfEntityQuery.list(RevisionsOfEntityQuery.java:120)   
        at org.hibernate.envers.query.internal.impl.AbstractAuditQuery.getSingleResult(AbstractAuditQuery.java:111)                                      
        at org.radarcns.management.service.RevisionService.getAuditInfo(RevisionService.java:113)                  
        at org.radarcns.management.service.mapper.decorator.UserMapperDecorator.userToUserDTO(UserMapperDecorator.java:31)
        at org.radarcns.management.web.rest.AccountResource.lambda$getAccount$1(AccountResource.java:94)           
        at java.util.Optional.map(Optional.java:215)                                     
        at org.radarcns.management.web.rest.AccountResource.getAccount(AccountResource.java:94)       
        at org.radarcns.management.web.rest.AccountResource$$FastClassBySpringCGLIB$$86cff444.invoke(<generated>)  
[...]

PR #441, opened 2019-05-31
Add health check

### LINKED ISSUE #440: Add a simple health check with database connection
Provide a simple health check endpoint with detection of database connection to application. E.g. 
https://www.onlinetutorialspoint.com/spring-boot/spring-boot-actuator-database-health-check.html

nivemaham:
Jhipster already added support for health-check using spring-actuator.
I have enabled the config to use database connection check and disable security checks for this specific call, to allow automatic health detection from docker/kubernetes health-check. 

The sample call and response are as follows:
```
curl 'http://localhost:9000/management/health' 
{
  "status" : "UP",
  "db" : {
    "status" : "UP",
    "database" : "H2",
    "hello" : 1
  },
  "refreshScope" : {
    "status" : "UP"
  },
  "hystrix" : {
    "status" : "UP"
  }
}
```

Fixes #440 #434

### DIFF
--- a/src/main/java/org/radarcns/management/config/OAuth2ServerConfiguration.java
+++ b/src/main/java/org/radarcns/management/config/OAuth2ServerConfiguration.java
@@ -131,6 +131,7 @@ public class OAuth2ServerConfiguration {
                     .antMatchers("/api/register").hasAnyAuthority(AuthoritiesConstants.SYS_ADMIN)
                     .antMatchers("/api/profile-info").permitAll()
                     .antMatchers("/api/**").authenticated()
+                    .antMatchers("/management/health").permitAll()
                     .antMatchers("/management/**").hasAnyAuthority(AuthoritiesConstants.SYS_ADMIN)
                     .antMatchers("/v2/api-docs/**").permitAll()
                     .antMatchers("/swagger-resources/configuration/ui").permitAll()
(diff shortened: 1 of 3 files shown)


k1-hedayati:
Does the endpoint return 5XX error when there is an error in database connection or other critical components?

src/main/java/org/radarcns/management/config/OAuth2ServerConfiguration.java
```diff
@@ -131,6 +131,7 @@ public void configure(HttpSecurity http) throws Exception {
                     .antMatchers("/api/register").hasAnyAuthority(AuthoritiesConstants.SYS_ADMIN)
                     .antMatchers("/api/profile-info").permitAll()
                     .antMatchers("/api/**").authenticated()
+                    .antMatchers("/management/health").permitAll()
```
blootsvoets: Could you add a comment here about why we're permitting all?

src/main/resources/config/application.yml
```diff
@@ -39,10 +39,15 @@ hystrix:
 management:
     security:
         roles: ROLE_SYS_ADMIN
+        enabled: false
     context-path: /management
     health:
+        diskspace:
```
blootsvoets: could you add a comment that we're disabling this because it is not authenticated and poses a small security issue?

PR #448, opened 2019-06-11
Pair app dialog

### LINKED ISSUE #447: Pair token manually dialog enhancements
The current pair client dialog has some minor usability issues:
1. The base URL should not contain the management portal URL, since it also refers to Kafka, etc.
2. The token uses a default font with hard to distinguish `I l 1` and `O 0`. I would propose to use a monospaced font like Inconsolata or Consolas to avoid reading errors.

blootsvoets:
Fixes #447.
- ~Use the configured base URL instead of the managementportal base url~
- Load the inconsolata font when the dialog is loaded. Do not load it more than once using a unique element ID.
- Specify that the token should use a monospaced font like Consolas, Inconsolata, Monaco or monospace.

### DIFF
--- a/src/main/java/org/radarcns/management/service/OAuthClientService.java
+++ b/src/main/java/org/radarcns/management/service/OAuthClientService.java
@@ -200,9 +200,11 @@ public class OAuthClientService {
 
         if (metaToken.getId() != null && metaToken.getTokenName() != null) {
             // get base url from settings
-            String baseUrl = managementPortalProperties.getCommon().getManagementPortalBaseUrl();
+            String baseUrl = managementPortalProperties.getCommon().getBaseUrl();
             // create complete uri string
-            String tokenUrl = baseUrl + ResourceUriService.getUri(metaToken).getPath();
+            String tokenUrl = managementPortalProperties.getCommon().getManagementPortalBaseUrl()
+                    + ResourceUriService.getUri(metaToken).getPath();
+
             // create response
             return new ClientPairInfoDTO(new URL(baseUrl), metaToken.getTokenName(),
                     new URL(tokenUrl), getMetaTokenTimeout());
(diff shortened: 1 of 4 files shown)


nivemaham:
As a base url in this case, we are considering the base-url of ManagementPortal which generates the token. This will involve changes on the apps. ManagementPortal Base URL is also configured based on the base-url on RADAR-Docker.

blootsvoets:
What's confusing though, is that the link of the QR code contains a URL that omits the `/managementportal/` url: https://github.com/RADAR-base/ManagementPortal/blob/574d50e142804d811a0968ceda6f73a22852f7eb/src/main/java/org/radarcns/management/service/MetaTokenService.java#L71. The Token URL is the same regardless... Indeed, https://github.com/RADAR-base/RADAR-Questionnaire/blob/4936ebd2a8b01aa2291e04c765902625c1669921/src/app/pages/auth/services/auth.service.ts#L105 would need to be changed to check whether `managementportal/` is already part of the URL or not, and to append it if needed.

nivemaham:
Does it? At least as of 5.5.0 , it contains the `/managementportal`
```json
{"tokenName":"WEFd1CNmKtrK","tokenUrl":"https://radarbase-demo.thehyve.net/managementportal/api/meta-token/WEFd1CNmKtrK","baseUrl":"https://radarbase-demo.thehyve.net/managementportal","timeout":"01:00","timesOutAt":"2019-06-11T16:09:44.431Z"}
```

nivemaham:
ok. I think there is a confusion. The Meta-token contains the base-url of the managment-portal.
The Token info contains, the base-url of the environment. Do we want to change the meta-token details not to contain the `managmentportal`? This might include a change on Thinc-IT, which is why I am little hesitant.

blootsvoets:
Ah yes you're right, that would use the same ClientPairDTO. Confusingly, the meta token itself contains a base URL without the managementportal...

blootsvoets:
OK, I've reverted that change.

PR #452, opened 2019-06-24
Reduce api results

### LINKED ISSUE #443: Project API /api/projects data duplication and performance
In the project DTO, there is a `sourceTypes` property, which has a `sourceData` property, which has again a `sourceType` property which duplicates information in the parent object. Could the child `sourceType` property be removed from `sourceData`, since it only duplicates data?

Also, would it be possible to remove the `sourceTypes` completely when querying for all projects, but rather reserve that for when a single project is being queried? Even in a single project, I'd think that `sourceData` is probably not necessary, only if `/api/projects/{projectId}/sourceTypes` is being queried.

blootsvoets:
Aims to fix #443.

Changes:
- If a source type is returned, its encapsulated `sourceData[].sourceType` is not returned.
- If a project is returned or users are listed as part of a project, project information is not returned
- If a project is returned as part of a user or subject, the project does not list source types. The only exception is a single subject request: here the source types are returned.
- Do not convert a subject to a DTO just for auth checking.

### DIFF
--- a/src/main/java/org/radarcns/management/service/ProjectService.java
+++ b/src/main/java/org/radarcns/management/service/ProjectService.java
@@ -64,7 +64,7 @@ public class ProjectService {
     public Page findAll(Boolean fetchMinimal, Pageable pageable) {
         Page<Project> projects = projectRepository.findAllWithEagerRelationships(pageable);
         if (!fetchMinimal) {
-            return projects.map(projectMapper::projectToProjectDTOWithoutSources);
+            return projects.map(projectMapper::projectToProjectDTOReduced);
         } else {
             return projects.map(projectMapper::projectToMinimalProjectDetailsDTO);
         }
--- a/src/main/java/org/radarcns/management/service/ProjectService.java
+++ b/src/main/java/org/radarcns/management/service/ProjectService.java
@@ -52,8 +52,7 @@ public class ProjectService {
         log.debug("Request to save Project : {}", projectDto);
         Project project = projectMapper.projectDTOToProject(projectDto);
         project = projectRepository.save(project);
-        ProjectDTO result = projectMapper.projectToProjectDTO(project);
-        return result;
+        return projectMapper.projectToProjectDTO(project);
     }
 
     /**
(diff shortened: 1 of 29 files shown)


src/main/java/org/radarcns/management/service/ProjectService.java
```diff
@@ -65,7 +64,7 @@ public ProjectDTO save(ProjectDTO projectDto) {
     public Page findAll(Boolean fetchMinimal, Pageable pageable) {
         Page<Project> projects = projectRepository.findAllWithEagerRelationships(pageable);
         if (!fetchMinimal) {
-            return projects.map(projectMapper::projectToProjectDTO);
+            return projects.map(projectMapper::projectToProjectDTOReduced);
```
nivemaham: The projectToMinimalProjectDetailsDTO also has part of the project data. Maybe we can remove the fetchMinimal condition and just use a single DTO?

src/main/java/org/radarcns/management/web/rest/SubjectResource.java
```diff
@@ -300,10 +301,17 @@
             throws NotAuthorizedException {
         log.debug("REST request to get Subject : {}", login);
         Subject subject = subjectService.findOneByLogin(login);
+        Project project = subject.getActiveProject()
+                .flatMap(p ->  projectRepository.findOneWithEagerRelationships(p.getId()))
+                .orElse(null);
+
+        String projectName = project == null ? null : project.getProjectName();
```
nivemaham: should we add InvalidStateException if the projectName is null?

src/main/java/org/radarcns/management/service/ProjectService.java
blootsvoets: Minimal project data only includes name and ID, but not the information needed to print the projects e.g. on the home page.

src/main/java/org/radarcns/management/web/rest/SubjectResource.java
blootsvoets: Is a project required for all subjects? If so, yes, it should be an invalid state. Otherwise, I don't see why a subject without a project should not be returned.

nivemaham: I think the subject can be returned without a project. If the subject is `discontinued`, I think then the active project becomes `null`. I was more concerned if the `checkPermissionOnSubject` can handle null for project. If so, it looks good

blootsvoets: I've added a some tests for this, indeed, `null` project authorization behaves as expected.

nivemaham:
@k1-hedayati Once the docker container for this is built after merge, would it be possible to test with the gatling script. This should fix the bottleneck with the queries, thanks to @blootsvoets .

k1-hedayati:
Great, Thanks @blootsvoets @nivemaham

PR #464, opened 2019-08-06
Add Jersey API calls for use with ManagementPortal

blootsvoets:
Some tweaks to radar-auth to simplify integration with radar-jersey.


radar-auth-jersey/README.md
```diff
@@ -0,0 +1,57 @@
+# radar-auth-jersey
```
nivemaham: Why can't we package it together with radar-auth? Is it due to the dependency to Jersey?

radar-auth-jersey/src/main/kotlin/org/radarbase/auth/jersey/Auth.kt
```diff
@@ -0,0 +1,25 @@
...
+package org.radarbase.auth.jersey
+
+import org.radarcns.auth.authorization.Permission
+
+interface Auth {
+    val clientId: String?
+    val defaultProject: String?
+    val userId: String?
+
+    fun hasPermissionOnProject(permission: Permission, projectId: String): Boolean
```
nivemaham: Maybe it is good to document these methods, to make it easier for the users.

radar-auth-jersey/src/main/kotlin/org/radarbase/auth/jersey/AuthConfig.kt
```diff
@@ -0,0 +1,21 @@
...
+
+package org.radarbase.auth.jersey
+
+data class AuthConfig(
+        val managementPortalUrl: String? = null,
+        val jwtResourceName: String,
+        val jwtIssuer: String? = null,
+        val jwtECPublicKeys: List<String>? = null,
+        val jwtRSAPublicKeys: List<String>? = null,
+        val jwtKeystorePath: String? = null,
```
nivemaham: Should we continue to support key-store related configs here, since after MP:0.5.4 key-store is related things are handled by MP. Having a config for retrieving JWK should sufficient, if I am not mistaken. Since It is a new library, it would be good to eliminate legacy configs.

radar-auth-jersey/src/main/kotlin/org/radarbase/auth/jersey/EcdsaJwtTokenValidator.kt
```diff
@@ -0,0 +1,117 @@
...
+import java.security.KeyStore
+import java.security.PublicKey
+import java.security.interfaces.ECPublicKey
+import java.security.interfaces.RSAPublicKey
+import java.security.spec.X509EncodedKeySpec
+import java.util.*
+import javax.ws.rs.container.ContainerRequestContext
+import javax.ws.rs.core.Context
+
+class EcdsaJwtTokenValidator constructor(@Context private val config: AuthConfig) : AuthValidator {
```
nivemaham: Now I am a little confused. Is this meant to replace the `radar-auth`? I see a lot of redundancy between `radar-auth` and this. My previous comment was based on the assumption that `radar-auth-jersey` would be a sub-library of `radar-auth`.

radar-auth-jersey/README.md
blootsvoets: Exactly, because of the Jersey dependency. If that's not a problem, we could also package it together.

radar-auth-jersey/src/main/kotlin/org/radarbase/auth/jersey/AuthConfig.kt
blootsvoets: These configs are only used if a raw ECDSA provider like Keycloak is used.

radar-auth-jersey/src/main/kotlin/org/radarbase/auth/jersey/EcdsaJwtTokenValidator.kt
blootsvoets: This validator is only used if a raw ECDSA provider like Keycloak is used.

blootsvoets: I have explicitly set Auth to extend JwtRadarToken now. This way, the relation between the authentication in radar-auth and radar-auth-jersey is more explicit.

blootsvoets:
@nivemaham can this still be merged as part of version 0.5.7?

nivemaham:
@blootsvoets ManagementPortal 0.5.7 has already been released. We can include this in the next release.

blootsvoets:
Moved radar-auth-jersey to a separate repo, it was too hard to get Kotlin + Java 11 in MP and test all projects together.

PR #476, opened 2019-10-24
allow project admin to create measurements.

nivemaham:
Part of Data uploader requirements


nivemaham:
> LGTM. In the longer term, it would be good if this was customisable, perhaps at the management portal?

Indeed. But would take some time.

PR #477, opened 2019-10-25
Fix navigating to user management

nivemaham:
- Fixes wrong implementation of Back button
- Adds additional check for admin route on the front-end
- Adds additional check to check for authority when requesting admin data on the back-end


nivemaham:
@yatharthranjan Could you confirm that PROJECT_ADMINs are not expected to visit the /user-managment/ view on ManagementPortal as of now?

yatharthranjan:
Hi guys, this should be fine. They are only supposed to work with subjects in the project view.

nivemaham:
@yatharthranjan Thanks yatharth

PR #487, opened 2019-12-17
Improve dependency injection and use correct OAuth token key endpoint

blootsvoets:
Dependency injection is now used for ManagementPortalOauthKeyStoreHandler. This allows it to get the local URL of the oauth path. It also simplifies the logic in other components for getting access to this handler.


blootsvoets:
Tested in a full radar-docker setup, this works now.

PR #494, opened 2020-05-06
Add persistent meta token

blootsvoets:
For submitting our app to the Google Play store, they require a valid login for possibly multiple app reviews. Since the meta token by default is valid for a short duration only and can only be used once, this is not an option. Moreover, for settings where in-person intervention is hard to do (e.g., because participant-researcher face-to-face contact is extremely sparse), it can be useful to give a QR code valid for a longer time.

For these reasons, this PR introduces the Persistent Meta Token. It is valid (by default) for 1 year and it can be reused. Changes made:
- added two symbols to token table to ensure that it has 64 options, making a character 6-bit equivalent.
- default meta tokens have 8 characters (=48 bits entropy), persistent meta tokens have 16 (=72 bits entropy).
- meta tokens can be deleted by users with the SUBJECT UPDATE permission for the subject the meta token was intended for.
- a refresh token is generated when a meta-token is fetched, not when it is stored. This makes token reuse easier to handle and avoids storing credentials in the database.
- frontend pairing dialog requires action to generate the QR code, as not too many QR codes should be active simultaneously. It deletes the previous QR codes if generate is called multiple times so only the last QR code is valid.
- frontend pairing dialog shows data table by default, including validity date and time.


yatharthranjan:
Yes, we had similar issues when submitting applications In the AppStore. But we just increased the token timeout on the dev instance as there is no real study data there. But, this looks very useful. Thanks

src/main/webapp/app/shared/subject/subject-pair-dialog.component.html
```diff
@@ -53,20 +62,5 @@ <h4 class="modal-title" id="mySubjectLabel" jhiTranslate="managementPortalApp.su
...
-                (click)="lockTokenUrl()">
-            <span class="fa fa-lock"></span>&nbsp;<span jhiTranslate="managementPortalApp.subject.hideTokenUrl">Hide Token Url</span>
-        </button>
-        <button type="button" class="btn btn-warning"
-                *ngIf="!showTokenUrl"
-                [disabled]="!showQRCode"
-                (click)="unlockTokenUrl()">
-            <span class="fa fa-unlock"></span>&nbsp;<span jhiTranslate="managementPortalApp.subject.showTokenUrl">Show Token Url</span>
-        </button>
-        <button type="button" class="btn btn-primary" [disabled]="!showQRCode" (click)="updateQRCode()">
```
nivemaham, line 68: The non persistent tokens can still be fetched only once. Wouldn't we want to have option to refresh if we choose to generate QR code? (not the persistent one)

blootsvoets: The Generate QR code button remains active, and can be pressed again to regenerate.

blootsvoets:
The same goes as with the existing token though: rate limiting is managed by the web server. The intrinsic security seems alright, since a much longer token is used. This approach avoids the situation where all meta-tokens are set to a long timeout, as @yatharthranjan indicates currently happens for the Google Play authentication. However, I agree that for some use cases the timeout could be managed perhaps on a per-project basis. If the default of 365 is too long, we could set it to a month instead? In any case, the default can be configured in the application properties.

With an earlier project as well, it seems like this would be a feature that would be nice to have: users can take their QR code and scan it after any server timeouts or issues or phone loss, instead of a researcher having to pay them a house visit.

Options:
1. leave as is
  - easiest for users to understand
  - may lead to inadvertently handing out persistent tokens where the project does not demand it
2. disable in configuration
  - layout will be easy for users to understand
  - the feature will be hard to discover
3. add a per-project checkbox to enable persistent tokens. All persistent tokens on the server will have the same timeout.
  - add some control to avoid handing out persistent tokens for projects that do not need it
  - potentially confuse users as to when the persistent token button is available
  - add some configuration overhead during project setup
4. add per-project persistent token checkbox and validity (days).
  - add some control to avoid handing out persistent tokens for projects that do not need it
  - add some control in relation to participant trial period and project duration
  - potentially confuse users as to when the persistent token button is available
  - add some configuration overhead during project setup
5. add per-token validity with a predefined maximum.
  - most flexible
  - adds configuration overhead for each token generated
  - may confuse users as to what value to put

I'm personally inclined to go for option 1. or 3. Alternatively, to just go for 1. now and make an issue to introduce 3. at a later time. @nivemaham @yatharthranjan what do you think?

nivemaham:
@blootsvoets I agree that this is a needed feature and the ability to request tokens for longer period is very useful. I am in favor of option 4 as well and the setting default to a month also sounds good. 
Splitting it into a separate issue sounds good to me.

blootsvoets:
Implemented option 2. for now, with the API ready to have option 4. as well.

PR #503, opened 2020-07-21
Fix Postgres connection error

blootsvoets:
Fixes HikariCP connection issues, caused by RevisionService using a single EntityManager. After the database becomes disconnected, that EntityManager is no longer valid and the whole management portal stops working.

Changes:
- Fix EntityManager in RevisionService
- Fix Dockerfile npm install step
- Actually use `postgresql_version` in gradle
- Include the `org.radarcns.management.repository` package when scanning, needed for auditing information.

Fix details:
Use `@Transactional` and `@PersistContext` to indicate that an valid EntityManager should always be supplied and that all operations should run in a transaction. DTO mappings are performed in the service to avoid leaking any database objects.


nivemaham (commented):
LGTM

PR #510, opened 2020-09-30
Add endpoint to get subjects by date created

mpgxvii:
- Add endpoint to get subjects that were created between specified `fromDate` and `toDate` for pulling MP data to Kafka


nivemaham:
This endpoint is only meant for SYS_ADMIN. What information do you need to be sent to Kafka?

mpgxvii:
> This endpoint is only meant for SYS_ADMIN. What information do you need to be sent to Kafka?

I was hoping to use revisions, since it is a stream of data with timestamps, to send subject data to Kafka (every time a subject is added in the revisions). Is this fine to expose to non `SYS_ADMIN`?

mpgxvii:
I've made the changes to allow pulling subjects by `createdDate`. This will then be polled by the connector. Is this fine?

src/main/java/org/radarcns/management/web/rest/SubjectResource.java
```diff
@@ -288,6 +290,27 @@
...
+    @GetMapping(value = "/subjects", params = { "fromDate", "toDate" })
+    public ResponseEntity<List<SubjectDTO>> getSubjectsByDateCreated(
+            @RequestParam(value = "fromDate")
+            @DateTimeFormat(iso = DateTimeFormat.ISO.DATE_TIME) ZonedDateTime fromDate,
+            @RequestParam(value = "toDate")
+            @DateTimeFormat(iso = DateTimeFormat.ISO.DATE_TIME) ZonedDateTime toDate,
+            @PageableDefault(page = 0, size = Integer.MAX_VALUE) Pageable pageable)
+            throws NotAuthorizedException {
+        log.debug("REST request to get subjects");
+        List<SubjectDTO> subjects = subjectService.findByDateCreated(fromDate, toDate);
```
blootsvoets, line 310: This list should definitely be filtered to only include subjects where token.hasPermissionOnSubject is true.

src/main/java/org/radarcns/management/service/SubjectService.java
```diff
@@ -205,6 +206,19 @@ public SubjectDTO updateSubject(SubjectDTO newSubjectDto) {
...
+     *
+     * @param fromDate date created from
+     * @param toDate date created to
+     * @return the requested list of subjects
+     */
+    public List<SubjectDTO> findByDateCreated(ZonedDateTime fromDate, ZonedDateTime toDate) {
+        return this.findAll(new PageRequest(0,Integer.MAX_VALUE)).getContent().stream()
+                .filter(subject -> subject.getCreatedDate().isAfter(fromDate)
+                && subject.getCreatedDate().isBefore(toDate)).collect(Collectors.toList());
+    }
```
blootsvoets, line 220: This should take a `Pageable` and return a `Page` instead of a list. Right now, the database query does not actually limit the dates, pages or authorization.

1. query should take pageable parameter and return a page.
2. query should include date parameters instead of using findAll (this needs an override in `SubjectRepository`)
3. query should include authorization parameters instead of filtering afterwards, otherwise the pagination will be off.

The following conditions are needed for authorization:
- before the request `token.checkPermission(SUBJECT_READ)` must be called.
- if the token is with client credentials OR the user is SYS_ADMIN, no limits on users are imposed.

In those cases, no filtering is needed. We could say we want to limit this API to OAuth clients and admins, in that case, we're done. All other types should get a forbidden.

Otherwise, you need to enumerate all projects where the token has role `PROJECT_*`. You can then create a query like `WHERE ... AND s.project.projectName IN ({$projectNames}). You would have to make a method for that in `SubjectRepository`.

mpgxvii:
Closing as this is not needed anymore, will re-open when needed.

PR #523, opened 2020-10-20
Ensure that refresh tokens from other resources can be read

blootsvoets:
Issue:
An OAuth client with audience other than `res_ManagementPortal` could not refresh its token. This occurs for example in frontend clients for a dedicated backend.

Cause:
JWTVerifiers, including those for refresh tokens, required that the audience of a JWT was `res_ManagementPortal`. This disabled the `refresh_token` OAuth 2.0 flow for clients that do not have `res_ManagementPortal` as an audience.

Fix:
First check whether a token is a refresh token. If so, don't check the audience. If it is an access token, the audience is still required to be `res_ManagementPortal`.


nivemaham (approved):
LGTM. Although the builds are failing. Can you have a look?

PR #545, opened 2021-01-04
Typeahead added for Project SourceType Selection

### LINKED ISSUE #539: Improve source selection UI when editing project
When editing a project, a multiselect box is shown with the available sources. When a lot of source types are present in the system, this is hard to use: it requires scrolling while selecting multiple sources, and you can only see a few sources at once. Preferably, this is implemented via an autocomplete box, and once an entry is selected it is added as a fixed label, with a delete icon next to it to remove it again.

aykl:
resolves #539 

https://user-images.githubusercontent.com/22550677/103534198-bb527d00-4e8e-11eb-9d3e-c7d930fa3b18.mov

### DIFF
--- a/src/main/webapp/app/entities/project/project-dialog.component.html
+++ b/src/main/webapp/app/entities/project/project-dialog.component.html
@@ -95,9 +95,27 @@
 
         <div class="form-group">
             <label jhiTranslate="managementPortalApp.project.sourceType" for="field_sourceType">Source Type</label>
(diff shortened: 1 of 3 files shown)


blootsvoets (approved):
LGTM

nivemaham (approved):
LGTM. For future, could we perhaps make this into a generic component, so that we can reuse this in other places?

Issue #530, opened 2020-11-02
Navigation in project tabs is not kept in history


blootsvoets:
When you go to sources, click a source and click back, it returns to 'Subjects' not 'Sources'. The URL stays the same for all tabs as well. Perhaps it would be possible to add information about the current tab and the current table page in the URL, e.g. `/managementportal/#/projects/radar-test/sources?page=2` Alternatively, `history.state` is used to store the state, so at least back and forward work. See e.g. https://developer.mozilla.org/en-US/docs/Web/API/History/pushState.

PR #547, opened 2021-01-07
Fix/pagination history

aykl:
Related to #530.
Navigation is tracked in history on _entity-specific pages_, but back/forward buttons did not cause page content to reload. This MR fixes it.

Original behavior:
https://user-images.githubusercontent.com/22550677/103892439-e4bc1480-50eb-11eb-9cbc-a6f00384b0d7.mov

Improved behavior:
https://user-images.githubusercontent.com/22550677/103892456-eab1f580-50eb-11eb-890a-333494beee5e.mov


blootsvoets (approved):
LGTM, cleans up some code and good catch on clarifying that `reversed` is actually `ascending` order...

PR #548, opened 2021-01-07
Upgrade gradle and plugins

blootsvoets:
Changes:
- Upgrades Gradle version, needed to use any new plugins
- Mandatory update to PMD 6.12.0
- Fixed any errors generated by the new PMD version

PR #551, opened 2021-01-11
Management portal client

blootsvoets:
Added a Kotlin ManagementPortal client. The build has been made compatible with Gradle 6 (includes #548) and updated _some_ plugins to be Java 11 compatible to allow for easier Kotlin integration.

Edit: clarify that not all plugins are Java 11 compatible.


managementportal-client/src/main/kotlin/org/radarbase/management/client/MPUser.kt
```diff
@@ -0,0 +1,32 @@
...
+ * See the file LICENSE in the root of this repository.
+ */
+
+package org.radarbase.management.client
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import java.time.ZonedDateTime
+
+data class MPUser(
+    @JsonProperty("login")
```
aykl, line 16: In `UserDTO` class I see two separate fields — `String login` and `Long id`. Shouldn't fields of `MPUser` correspond to those of `UserDTO`?

blootsvoets: Indeed. Well for purposes of external clients, the ID of the user is meaningless: it's an internal database identifier not used elsewhere. For all intents and purposes, the login acts as an ID instead. By converting the name, the intent of the login field as identifier becomes clearer in my opinion.

managementportal-client/src/main/kotlin/org/radarbase/management/client/MPClient.kt
```diff
@@ -0,0 +1,195 @@
...
+import org.radarcns.oauth.OAuth2Client
+import java.io.IOException
+import java.net.MalformedURLException
+import java.util.concurrent.TimeUnit
+
+/**
+ * Client for the ManagementPortal REST API.
+ */
+@Suppress("unused")
+class MPClient private constructor(
```
aykl, line 34: Readme suggests using this constructor, but it is private. How would it work?

aykl (approved):
L Mostly* GTM.

[*] It builds and passes tests with JDK 11, but `bootRun` results in an exception (starting with `class jdk.internal.loader.ClassLoaders$AppClassLoader cannot be cast to class java.net.URLClassLoader`). Tests fail with JDK 8, but the project runs as expected. I see it as an acceptable, albeit inconvenient, transitional situation, but it should either be fixed soon, or noted somewhere in readme/issues/other known place.

blootsvoets:
Re: Java 11, yes I didn't realise bootRun depended on Java 8. So this project needs to be updated to Spring Boot 2 to function under Java 11, quite a big migration. Fortunately, I've discovered that the Kotlin project also works with Java 8.

PR #553, opened 2021-01-19
Fix permissions for actually removing meta tokens

blootsvoets:
Deleting the meta-tokens did not work correctly, it would always throw forbidden.


aykl (approved):
LGTM.

PR #554, opened 2021-01-21
Fixed references to other dependencies

blootsvoets:
Apparently, shadowJar also renames dependencies that are not directly packaged. This PR fixes that by only relocating the relevant namespaces and none other.


yatharthranjan (approved):
LGTM

PR #555, opened 2021-01-27
Continue with other verifiers on SignatureVerificationException

yatharthranjan:
This allows public keys with different signatures to be used (for example from different MP deployments)

### DIFF
--- a/radar-auth/src/main/java/org/radarcns/auth/authentication/TokenValidator.java
+++ b/radar-auth/src/main/java/org/radarcns/auth/authentication/TokenValidator.java
@@ -150,10 +150,9 @@ public class TokenValidator {
                 }
                 return new JwtRadarToken(jwt);
             } catch (SignatureVerificationException sve) {
-                LOGGER.warn("Client presented a token with an incorrect signature, fetching public "
-                        + "keys again. Token: {}", token);
+                LOGGER.debug("Client presented a token with an incorrect signature. Token: {}",
+                        token);
                 refresh();
-                return validateAccessToken(token);
             } catch (JWTVerificationException ex) {
                 LOGGER.debug("Verifier {} with implementation {} did not accept token {}",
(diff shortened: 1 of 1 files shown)


radar-auth/src/main/java/org/radarcns/auth/authentication/TokenValidator.java
```diff
@@ -150,10 +150,9 @@ public RadarToken validateAccessToken(String token) throws TokenValidationExcept
                 }
                 return new JwtRadarToken(jwt);
             } catch (SignatureVerificationException sve) {
-                LOGGER.warn("Client presented a token with an incorrect signature, fetching public "
-                        + "keys again. Token: {}", token);
+                LOGGER.debug("Client presented a token with an incorrect signature. Token: {}",
```
blootsvoets, line 153: Perhaps it would be best to remove the token altogether: it leaks credentials to the logs.

radar-auth/src/main/java/org/radarcns/auth/authentication/TokenValidator.java
```diff
@@ -150,10 +150,9 @@ public RadarToken validateAccessToken(String token) throws TokenValidationExcept
...
+                        token);
                 refresh();
-                return validateAccessToken(token);
```
blootsvoets, line 156: This should set a boolean: if keys were refreshed but no validator validated the token, maybe they will succeed with the new keys. This retry cycle can happen only once.

blootsvoets: Alternatively, add a private method with an additional “tryRefresh” boolean. This will be true in the first call, and false in the recursive call.

blootsvoets (approved):
LGTM, thanks for fixing the migration flow!

PR #562, opened 2021-02-10
Fix subject revision list not showing

mpgxvii:
- Fixes front end issue (even when backend sends correct results, the front end isn't showing the revisions)
- Simplify subject revisions endpoint

Aims to solve  #558


src/main/java/org/radarcns/management/web/rest/SubjectResource.java
```diff
@@ -319,32 +319,22 @@
...
             @ApiParam Pageable pageable, @PathVariable String login) throws NotAuthorizedException {
         log.debug("REST request to get revisions for Subject : {}", login);
-        SubjectDTO subject = subjectService.getLatestRevision(login);
-        Page<RevisionDTO> page = revisionService.getRevisionsForEntity(pageable,
-                subjectMapper.subjectDTOToSubject(subject));
-        SubjectDTO blank = new SubjectDTO();
-        RadarToken token = getJWT(servletRequest);
-        // iterate the revisions, if we don't have read permission for the subject in a given
-        // revision, swap the subject with a 'blank' subject so no information gets leaked
-        page = page.map(rev -> token.hasPermissionOnSubject(SUBJECT_READ,
```
mpgxvii, line 336: @yatharthranjan @blootsvoets @nivemaham Would it be alright to remove this mapping and just check for permissions of the subject once since the revision results are only the specified subject's revisions (in `getRevisionsForEntity`)?

blootsvoets: It could be that this is due to subjects moving from one project to another. We generally don't do this, so it might be fine to remove it...

yatharthranjan (dismissed):
Looks fine to me, but probably Joris or Nivethika should review before merging.

src/main/java/org/radarcns/management/web/rest/SubjectResource.java
```diff
@@ -326,25 +326,15 @@
...
-                .map(RevisionDTO::getEntity)
-                .allMatch(blank::equals)) {
-            throw new NotAuthorizedException();
-        }
+        Subject subject = subjectService.findOneByLogin(login);
+        Project project = subject.getActiveProject()
+                .flatMap(p -> projectRepository.findOneWithEagerRelationships(p.getId()))
+                .orElse(null);
+
+        checkPermissionOnSubject(getJWT(servletRequest), SUBJECT_READ, project.getProjectName(),
```
blootsvoets, line 334: This will throw NPE if project is null. better to throw intentionally in .orElse. I’m not sure about the implementation of checkpermissiononsubject if the project name is null. In that case, only a user or client with global level access should be able to see the data.

src/main/java/org/radarcns/management/web/rest/SubjectResource.java
```diff
@@ -326,25 +326,15 @@
...
-        // access to the requested subject in any of the subject history's projects from gaining
-        // information on the subject history this way, we throw a NotAuthorized.
-        if (page.getContent().stream()
-                .map(RevisionDTO::getEntity)
-                .allMatch(blank::equals)) {
-            throw new NotAuthorizedException();
-        }
+        Subject subject = subjectService.findOneByLogin(login);
+        Project project = subject.getActiveProject()
+                .flatMap(p -> projectRepository.findOneWithEagerRelationships(p.getId()))
```
blootsvoets, line 331: You could add a .map(p -> p.getProjectName()) here.

blootsvoets (approved):
LGTM

PR #564, opened 2021-02-11
Separate npm installs in Dockerfile

mpgxvii:
- The Docker build fails with the error:```The command '/bin/sh -c curl -sL https://deb.nodesource.com/setup_12.x | bash - && apt-get install -yq nodejs build-essential && npm install -g npm yarn' returned a non-zero code: 1```
- Separating the npm commands fixes it


blootsvoets (approved):
LGTM

PR #565, opened 2021-02-15
Ensure that only strong passwords are allowed

blootsvoets:
This changes the required 4 (!) characters to have at least 8 characters with a mix of alphanumeric characters with symbols. This matches the password policy of many established services.


nivemaham (dismissed):
LGTM. Question: Why do we check for password strength on both client and server side? Is the server side is implemented as an additional measure to ensure that passwords meet the requirements?

blootsvoets:
Indeed, we don't want clients to be able to circumvent password requirements by using the REST API.

aykl (approved):
LGTM.

PR #569, opened 2021-03-02
Fix failure when refresh keys too soon

yatharthranjan:
When too many requests(having a token that throws `SignatureVerificationException` on the first validator) are made too soon (<`fetchTimeout`), then the token validator fails with
 
```
401 org.radarcns.auth.exception.TokenValidationException: Not fetching public key more than once every PT1M
```

This PR ensures it validates the tokens even when refreshing of public keys fails (in our case due to requesting refresh too soon).


blootsvoets (approved):
LGTM

PR #571, opened 2021-03-08
Data Class Properties made Public

aykl:
Private properties make it impossible to update to the latest Radar-Jersey.
E.g. https://github.com/RADAR-base/radar-upload-source-connector/blob/master/radar-upload-backend/src/main/java/org/radarbase/upload/dto/Project.kt#L42 — `MPUser#id` can not be accessed.


blootsvoets (approved):
LGTM, thanks for taking this up.

PR #572, opened 2021-03-08
Use Maven Central for publishing artefacts

blootsvoets:
- Changes the groupId and base package from `org.radarcns` to `org.radarbase`
- Publishes the packages to OSSRH (Maven Central) instead of Bintray and artifactory.


yatharthranjan (approved):
LGTM.

PR #574, opened 2021-03-22
Better assigned source deletion error

aykl:
This fix does not implement any source archival procedures, but it gives a user a better idea of what exactly went wrong.
A proper fix requires more design effort.


blootsvoets (approved):
LGTM

PR #588, opened 2021-04-19
Source type selection fix

### LINKED ISSUE #587: The Source Type box is not populated with the selected dropdown value
Administrator -> Projects -> Create new project -> Select source type. Then `[object Object]` will be displayed rather than the true selected value:

source_type

aykl:
When a new Project is created, the `project` instance had uninitialized `sourceTypes` field leading to incorrect behavior. Initializing the field when it is edited resolves #587.


aykl:
Closing as Github doesn't update this PR with the latest commit.

PR #590, opened 2021-04-23
Update docs about paths to keystore.p12 and make the expiration consistent with application-*.yml

baixiac:
Update the paths to keystore.p12 and make the expiration consistent with application-*.yml.


.gitignore
```diff
@@ -7,7 +7,8 @@ generated/
 /src/test/javascript/coverage/
 /src/test/javascript/PhantomJS*/
 keystore.jks
-keystore.p12
```
blootsvoets, line 10: just `keystore.p12` is intentional. Any file that should then not be ignored is intentional.

blootsvoets (commented):
One comment, fine to merge otherwise.

.gitignore
baixiac: Oops... Reverted my unintentional change.

blootsvoets (approved):

Issue #424, opened 2019-03-20
Migrate to Spring 2.X


nivemaham:
Assuming we want to keep maintaining MP, it is important to migrate to Spring-boot 2.0, to keep us up to data and supported. 
This could be a starting point .https://dzone.com/articles/spring-boot-migration-from-15-to-205-release

PR #591, opened 2021-05-04
Upgrade to Spring Boot 2.3, Spring Core 5.2 and JUnit 5

baixiac:
...Of course, lots of other things could be upgraded as well but this PR is intended to keep the change minimum.

Closes https://github.com/RADAR-base/ManagementPortal/issues/424.


blootsvoets (commented):
That's a sizeable update, thanks for this! Does the authentication code still work as expected (to test with MP website, client credentials curl request, and another website, e.g. rest-source-authorizer)?

baixiac:
Hi, @blootsvoets , @mpgxvii and I are still working on the authorization_code flow and will let you know when that is finished.

baixiac:
Hi, @blootsvoets, please consider merging this after review. Thank @mpgxvii for testing it with rest-source-authorizer.

blootsvoets:
On boot, I get the following warning:
```
WARN 1 --- [           main] o.s.boot.actuate.endpoint.EndpointId     : Endpoint ID 'hystrix.stream' contains invalid characters, please migrate to a valid format.
```

It is probably related to this change:
https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.1-Release-Notes#endpoint-id-names

Could you take a look on whether the default name can be changed to fix this warning?

blootsvoets:
The spring security modules used seem deprecated according to this message:
```
WARN 1 --- [           main] s.o.SpringCloudSecurityAutoConfiguration : All Spring Cloud Security modules and starters are deprecated. They will be moved to individual projects in the next major release.
```
Is this something that should be fixed, or can it be postponed?

blootsvoets:
I've tested the deployment, LGTM. Could you take a look at the two warning messages? If those are addressed or made into separate issues, I'm happy to merge.

baixiac:
Well spotted, @blootsvoets . Will have a look. Looks like hystrix is in maintenance mode only and maybe getting rid of it is the ultimate solution.

baixiac:
Hi, @blootsvoets . Looks the previous version has got a bunch of warnings already. Given the regression test such as the rest-source-authorizer part is not cheap atm, shall we create a separate task for getting rid of all the warnings?

blootsvoets (approved):

PR #600, opened 2021-05-11
Fix TemplateInputException in login template

mpgxvii:
- Getting an error `org.thymeleaf.exceptions.TemplateInputException: An error happened during template parsing (template: "mails/login.html")` when opening `/managementportal/login` endpoint
- Although default thymeleaf directory is `templates/`, all templates are resolving to the `mails/` directory
- This moves mail templates to `templates/` directory so all templates are in the same directory

Depends on #591


baixiac:
Nice fix! 👍 
I was wondering why there were 404s on `/mails`'s error page so I copied it across from `/templates`.

PR #605, opened 2021-05-18
Make management portal run on Java 11

### LINKED ISSUE #458: Support Java 11
Right now the gatling scripts and possibly some other parts do not run or even compile with Java 11/12. This should be supported.

baixiac:
Closes #458.


baixiac:
@mpgxvii kindly conducted another round of regression test with radar-source-auth on this change.

Dockerfile
```diff
@@ -34,7 +34,7 @@ COPY src src
 RUN ./gradlew -s bootWar
 
 # Run stage
-FROM openjdk:8-jre-alpine
+FROM adoptopenjdk/openjdk11:alpine-jre
```
blootsvoets, line 37: For our other images we use `openjdk:11-jre-slim`, could that be replicated here?

build.gradle
```diff
@@ -53,8 +53,8 @@ allprojects {
             downloadSources = true
         }
     }
-    sourceCompatibility = '1.8'
-    targetCompatibility = '1.8'
+    sourceCompatibility = "11"
```
blootsvoets, line 56: Library source and target compatibility should remain 1.8 for radar-auth, since it’s used as a library potentially in Java 8 projects. For the root project, Java 11 is fine, as well as for managementportal-client: all kotlin projects can be expected to be migrated to Java 11 by now.

baixiac: That makes sense. Didn't know radar-auth is a shared lib. Change reverted.

In that case, should radar-auth be better extracted into its own codebase instead of being inside this mono-repo? Of cause, in a separate issue.

Dockerfile
baixiac: Done.

build.gradle
blootsvoets: The authorization library is closely tied to managementportal, so we'd like to keep them together. An improvement to the repository structure could be to make the managementportal root module a submodule as well, to clarify the relationship between the modules.

blootsvoets (dismissed):
LGTM! Thanks for this update.

build.gradle
blootsvoets: Indeed, perhaps take out sourceCompatibility out of the `allprojects` block and add it manually to each project with the appropriate version:
radar-auth: 1.8
radar-auth/deprecated-auth0: 1.8
root: 11
managemeåtportal-client: 11
oauth-client: 1.8

baixiac: Yes, let's do that.

blootsvoets (approved):

PR #606, opened 2021-05-19
Do not use enums for API data: there may be a mismatch

blootsvoets:
This avoids a situation where a mismatch does not import the source types.


src/main/java/org/radarbase/management/config/SourceTypeLoader.java
```diff
@@ -134,21 +134,29 @@ public void saveSourceTypesFromCatalogServer(List<CatalogSourceType> catalogSour
...
+                                    + "_" + sourceType.getCatalogVersion()
+                                    + "_" + sourceData.getSourceDataType());
+                            sourceData.sourceType(sourceType);
+                            sourceDataRepository.save(sourceData);
+                        } catch (RuntimeException ex) {
+                            log.error("Failed to import source data {}", catalogSourceData, ex);
+                        }
+                    }
+                } catch (RuntimeException ex) {
+                    log.error("Failed to import source type {}", sourceType, ex);
```
baixiac, line 159: Looks like `catch`es are only used for logging here. Not sure about the impact given the exceptions are swallowed.

baixiac (commented):
LGTM. One question about the error handling.

nivemaham (approved):

src/main/java/org/radarbase/management/config/SourceTypeLoader.java
blootsvoets: The exception swallowing is intentional. The previous behaviour was that exceptions were swallowed one level up, but then registration just stopped completely at the malformed source type.

baixiac: 👍

PR #608, opened 2021-05-24
Fix the retrieval on revision instant

baixiac:


yatharthranjan (dismissed):
LGTM

src/main/java/org/radarbase/management/domain/audit/CustomRevisionMetadata.java
```diff
@@ -36,7 +36,10 @@ public Integer getRequiredRevisionNumber() {
 
     @Override
     public Optional<Instant> getRevisionInstant() {
-        return Optional.empty();
+        if (entity.getTimestamp() == null) {
```
blootsvoets, line 39: Good find. Alternative using optional chaining:

```java
return Optional.ofNullable(entity.getTimestamp())
        .map(t -> t.toInstant());
```

baixiac: 👍  Improvement done.

blootsvoets (approved):

PR #615, opened 2021-08-09
Frontend upgrade

aykl:
- Translations switched to '@ngx-translate';
- 'ng-jhipster' dependency removed;
- Missing translations added;
- Angular version updated;
- Unused (e.g. Registration) and inaccessible (e.g. Spring Config page) components removed;
- Explicitly registered services (i.e. services declared in the `provider` module field) converted into implicitly-available singletons (using `providedIn: 'root'`, as recommended single Angular 6);


blootsvoets:
On `yarn install`, I get the following warnings:
```
warning " > angularx-qrcode@11.0.0" has incorrect peer dependency "@angular/core@^11.0.6".
warning " > bootstrap@4.6.0" has unmet peer dependency "popper.js@^1.16.1".
warning " > merge-json-webpack-plugin@6.0.0" has unmet peer dependency "webpack@^5.0.0".
```
Are these expected or can they be solved?

blootsvoets:
Looks good to me. Only part that I'm not sure about is the Administration -> Health page. The API gives back some health metrics but the table is empty.

aykl:
Yarn warnings are expected.
> `angularx-qrcode@11.0.0" has incorrect peer dependency "@angular/core@^11.0.6`
They are compatible, and once [the next version is released](https://github.com/Cordobo/angularx-qrcode/pull/132/files) it will be easy to make this warning go away.
> `bootstrap@4.6.0" has unmet peer dependency "popper.js@^1.16.1`
We don't use any components relying on popper, but I'll add this dependency.
> `merge-json-webpack-plugin@6.0.0" has unmet peer dependency "webpack@^5.0.0`
I decided not to specify webpack as an explicit dependency, since once angular is installed webpack is guaranteed to be installed as well. But yarn is not satisfied with transitive dependencies, hence we see the warning (which I intend to ignore).

src/main/webapp/app/admin/audits/audits.component.ts
```diff
@@ -59,7 +56,7 @@ export class AuditsComponent implements OnInit {
             fromDate: this.fromDate, toDate: this.toDate,
         }).subscribe((res: HttpResponse<any>) => {
             this.audits = res.body;
-            this.links = this.parseLinks.parse(res.headers.get('link'));
+            this.links = parseLinks(res.headers.get('link'));
```
nivemaham, line 59: I noticed that the audit entries are not shown on the UI, although I do not see any notable changes that could have caused it. Maybe it is not a side effect from this PR. But worth looking at.

src/main/webapp/i18n/en/sourceType.json
```diff
@@ -15,6 +15,8 @@
             "detail": {
                 "title": "Source Type"
             },
+            "canRegisterDynamically": "Dynamic Source Registration",
```
nivemaham, line 18: Don't we need the dutch version as well?

src/main/webapp/i18n/en/global.json
```diff
@@ -56,12 +56,7 @@
             "info": {
                 "authenticated": {
                     "prefix": "Are you a registered user? ",
-                    "link": "Sign in",
-                    "suffix": ", you can try the default accounts:<br/>- Administrator (login=\"admin\" and password=\"admin\") <br/>- User (login=\"user\" and password=\"user\")."
-                },
-                "register": {
-                    "noaccount": "You don't have an account yet?",
-                    "link": "Register a new account"
+                    "link": "Sign in"
```
nivemaham, line 59: Can you please change the link in the `footer` with the correct link https://radar-base.atlassian.net/wiki/spaces/RAD/pages/1913978881/User+Guide+for+Management+Portal,

src/main/webapp/app/admin/audits/audits.component.ts
aykl: I think it wasn't caused by this PR, but I've fixed it.

src/main/webapp/i18n/en/global.json
aykl: Done.

nivemaham: Thank you :+1:

nivemaham (dismissed):
Such a big effort. Thanks for getting this done.
I noticed a few bugs. But I don't think they are caused by this. 
@aykl Can you double check the build failures and resolve the merge conflict on Dockerfile? 
Otherwise, Looks good to me

nivemaham:
@aykl One thing I forgot to mention. Can you update the README with the recent commands where necessary?

aykl:
The only remaining red check comes from Snyk, but it can not be addressed before an updated `angularx-qrcode` module is released. However, the reported CVE can not be exploited in this app. Is there a way to ignore this particular Snyk check?

> @aykl One thing I forgot to mention. Can you update the README with the recent commands where necessary?

The README is up-to-date, old commands are still applicable, no changes required.

nivemaham:
@aykl We can proceed with the merge and update the qrcode lib once it is released.

nivemaham (approved):
LGTM

PR #617, opened 2021-09-01
Misc improvements

nivemaham:


src/main/java/org/radarbase/management/web/rest/ProjectResource.java
```diff
@@ -261,7 +262,12 @@ public ResponseEntity getAllSourcesForProject(@ApiParam Pageable pageable,
                     Boolean minimized) throws NotAuthorizedException {
         log.debug("REST request to get all Sources");
         ProjectDTO projectDto = projectService.findOneByName(projectName);
-        checkPermissionOnProject(getJWT(servletRequest), SOURCE_READ, projectDto.getProjectName());
+        RadarToken jwt = getJWT(servletRequest);
+        checkPermissionOnProject(jwt, SOURCE_READ, projectDto.getProjectName());
+        if (jwt.hasAuthority(PARTICIPANT)) {
+            throw new NotAuthorizedException(String.format("Client %s does not have "
```
nivemaham, line 268: Would it be better to rephrase the mesage that a PARTICIPANT cannot perform this operation. We are not checking for permission here?

src/main/java/org/radarbase/management/web/rest/RoleResource.java
```diff
@@ -135,6 +138,7 @@
     @GetMapping("/roles/{projectName:" + Constants.ENTITY_ID_REGEX + "}/{authorityName:"
             + Constants.ENTITY_ID_REGEX + "}")
     @Timed
+    @Secured({AuthoritiesConstants.PROJECT_ADMIN, AuthoritiesConstants.SYS_ADMIN})
```
nivemaham, line 141: I think except this call, all the others should be limited to SYS_ADMIN only.

src/main/java/org/radarbase/management/web/rest/SourceDataResource.java
```diff
@@ -71,7 +73,7 @@
     public ResponseEntity<SourceDataDTO> createSourceData(@Valid @RequestBody SourceDataDTO
             sourceDataDto) throws URISyntaxException, NotAuthorizedException {
         log.debug("REST request to save SourceData : {}", sourceDataDto);
-        checkPermission(getJWT(servletRequest), SOURCEDATA_CREATE);
+        checkAuthorityAndPermission(getJWT(servletRequest), SYS_ADMIN, SOURCEDATA_CREATE);
```
nivemaham, line 76: Just to confirm, would this disregard if the action is performed by a service with client_credentials?

src/main/java/org/radarbase/management/web/rest/SourceResource.java
```diff
@@ -77,7 +85,9 @@
     public ResponseEntity<SourceDTO> createSource(@Valid @RequestBody SourceDTO sourceDto)
             throws URISyntaxException, NotAuthorizedException {
         log.debug("REST request to save Source : {}", sourceDto);
-        checkPermission(getJWT(servletRequest), SOURCE_CREATE);
+        MinimalProjectDetailsDTO project = sourceDto.getProject();
+        String projectName = project != null ? project.getProjectName() : null;
+        checkPermissionOnProject(getJWT(servletRequest), SOURCE_CREATE, projectName);
```
nivemaham, line 90: This means a source cannot be created without being assigned to a project? Functionally sounds good to me. We may have to make the project required on the UI under entities/sources/create new source (until we remove that feature).

src/main/java/org/radarbase/management/service/SourceService.java
```diff
@@ -208,24 +212,41 @@ public MinimalSourceDetailsDTO safeUpdateOfAttributes(Source sourceToUpdate,
...
+        Source existingSource = existingSourceOpt.get();
+        String project = existingSource.getProject() != null
+                ? existingSource.getProject().getProjectName()
+                : null;
+        String user = (existingSource.getSubject() != null
+                && existingSource.getSubject().getUser() != null)
+                ? existingSource.getSubject().getUser().getLogin()
+                : null;
+
+        checkPermissionOnSource(jwt, SOURCE_UPDATE, project, user, existingSource.getSourceName());
```
nivemaham, line 234: Did we already have these methods (checkPermissionOnSource and checkPermissionOnSubject) available in the radar-auth?

nivemaham (commented):
I can't approve the PR myself. But mostly looks good to me. I left some minor comments and clarification questions.

radar-auth/src/main/java/org/radarbase/auth/token/AbstractRadarToken.java
```diff
@@ -84,10 +85,14 @@ protected boolean hasAuthorityForPermission(Permission permission) {
      * @return {@code true} if any authority contains the permission, {@code false} otherwise
      */
     protected boolean hasAuthorityForProject(Permission permission, String projectName) {
-        return getRoles()
-                .getOrDefault(projectName, emptyList()).stream()
-                .anyMatch(permission::isAuthorityAllowed)
-                || hasNonProjectRelatedAuthorityForPermission(permission);
+        if (hasNonProjectRelatedAuthorityForPermission(permission)) {
```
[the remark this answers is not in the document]
blootsvoets: Indeed, it's only called internally by `hasPermission.*` which checks for client credentials already.

src/main/java/org/radarbase/management/service/SourceService.java
blootsvoets: These have been available quite some time, but they were introduced later than the other permission checks.

src/main/java/org/radarbase/management/web/rest/SourceDataResource.java
blootsvoets: Good find. It didn't, but it does now.

src/main/java/org/radarbase/management/web/rest/SourceResource.java
blootsvoets: If the project is not valid (like `null`), `checkPermissionOnProject` will only allow sys admins and oauth clients. That is fine for this use case.

blootsvoets (approved):
Works -- thanks!

PR #621, opened 2021-09-06
Docker flow

blootsvoets:
Adds Docker build to Github Actions. Also builds docker images for Github releases, since Docker hub will no longer do that.


nivemaham (approved):
LGTM

PR #626, opened 2021-10-21
Full dependency update

blootsvoets:
Description: Update dependencies
- Gradle version
- Node plugin version (com.mooworks.node is no longer supported.)
- Spring Boot 2.5
- Spring Security 5.3
- Hazelcast 4
- Hibernate 5.4
- io.github.jhipster -> tech.jhipster
- Spring Session
- Removed deprecated Zuul Proxy.
- OpenAPI 3.0
- Gatling 3.6
- Node 16
- Yarn 2
- Removed unused dependencies 

#### Checklist:


.github/workflows/main.yml
```diff
@@ -22,14 +21,14 @@ jobs:
...
+      - name: Use Node.js 16
         uses: actions/setup-node@v1
         with:
-          node-version: 15
+          node-version: 16
 
       - name: Get yarn cache directory path
         id: yarn-cache-dir-path
-        run: echo "::set-output name=dir::$(yarn cache dir)"
+        run: echo "::set-output name=dir::.yarn/cache"
```
aykl, line 31: `pre-release.yml` and `release.yml` use `.yarn` directory. Is this discrepancy intentional?

blootsvoets: `.yarn/cache` is correct - fixing now.

src/test/javascript/e2e/cypress/integration/util/login.ts
```diff
@@ -1,16 +1,13 @@
...
-        },
-        body: `username=${username}&password=${password}&grant_type=password`
-    }).then(res => {
-        let token = res.body;
-        const expiredAt = new Date();
-        expiredAt.setSeconds(expiredAt.getSeconds() + token.expires_in);
-        token.expires_at = expiredAt.getTime();
-        cy.setCookie("oAtkn", JSON.stringify(token));
-    });
+    cy.visit('./');
```
aykl, line 4: This is the "right" way to log in, but it is slower. Previous version intentionally sends a request instead of signing in via the UI to speed up the tests that do not verify the Log In functionality.
(I do not see this change as an improvement)

blootsvoets: A few ideas here:
- the old way of logging in no longer worked: without Zuul proxy a session needs to be used, not a token. This way makes e2e logins behave as they would in the frontend, and do not rely on the backend implementation of tokens
- decouple tests
- I started using the cy.session API to speed up logins, but it is in beta and proved not to be reliable yet.

as a workaround, I could login once per test file and preserve the `SESSION` cookie. It would make the coupling between tests slightly stronger though.

aykl: >without Zuul proxy a session needs to be used, not a token
Okay, I didn't realize this was the case, I thought it was an "unprovoked" change.

No objections then. I didn't run the tests yet — maybe they are fast enough as-is with the current approach.

blootsvoets: I just checked; it used to be 4:50 min, it’s 4:35 min now - perhaps because of some source data tests that had been commented out though.

src/main/webapp/app/blocks/interceptor/auth.interceptor.ts
```diff
@@ -1,28 +1,28 @@
...
-                        Authorization: 'Bearer ' + token.access_token
-                    }
-                });
-                return next.handle(newReq);
-            }
-        }
-        return next.handle(request);
-    }
-}
+// import { Injectable } from '@angular/core';
```
aykl, line 1: I believe it can be removed.

aykl (approved):
LGTM.

PR #627, opened 2021-11-01
Enrollment date script

blootsvoets:
Description:
Ensure that the subject enrolment date matches its created date when the enrolment date variable is introduced.

### DIFF
--- a/src/main/resources/config/liquibase/changelog/20211006170000_add_enrollment_date.xml
+++ b/src/main/resources/config/liquibase/changelog/20211006170000_add_enrollment_date.xml
@@ -41,7 +41,6 @@
                     INNER JOIN _revisions_info r ON a.rev = r.id
                 WHERE a.id = subject.id
                     AND a.revtype = 0
-                GROUP BY a.id
                 )
             WHERE subject.enrollment_date IS NULL
         </sql>
--- a/src/main/resources/config/liquibase/changelog/20211006170000_add_enrollment_date.xml
+++ b/src/main/resources/config/liquibase/changelog/20211006170000_add_enrollment_date.xml
@@ -24,10 +24,26 @@
             <column name="enrollment_date" type="timestamp"/>
         </addColumn>
     </changeSet>
+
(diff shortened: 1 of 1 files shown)


src/main/resources/config/liquibase/changelog/20211006170000_add_enrollment_date.xml
```diff
@@ -24,10 +24,26 @@
...
+    <changeSet id="add subject enrollment date" author="joris@thehyve.nl">
+        <sql>
+            UPDATE subject
+            SET subject.enrollment_date = (
+                SELECT MIN(r.timestamp)
+                FROM subject_aud a
+                    INNER JOIN _revisions_info r ON a.rev = r.id
+                WHERE a.id = subject.id
+                    AND a.revtype = 0
+                GROUP BY a.id
```
aykl, line 44: Isn't it redundant? Aggregate functions like `MIN` return one result even without `GROUP BY`.

src/main/resources/config/liquibase/changelog/20211006170000_add_enrollment_date.xml
```diff
@@ -24,10 +24,26 @@
...
     </changeSet>
+
     <changeSet id="20211006170000-2" author="alexey@thehyve.nl">
         <addColumn tableName="subject_aud">
             <column name="enrollment_date" type="timestamp"/>
             <column name="enrollment_date_mod" type="BOOLEAN"/>
         </addColumn>
     </changeSet>
+
+    <changeSet id="add subject enrollment date" author="joris@thehyve.nl">
```
aykl, line 35: I like this id. It makes me wonder why don't we identify other changesets is the same readable manner.

aykl:
LGTM.

src/main/resources/config/liquibase/changelog/20211006170000_add_enrollment_date.xml
blootsvoets: I'm not sure either :-) I noticed this type of ID being used elsewhere and it's much more readable, also if looking through the liquidase change log table.

PR #628, opened 2021-11-03
Feature/add group to subjects

aykl:
PATCH request to add multiple subjects to a group in one API call.


blootsvoets (commented):
Good general structure. For the updateGroupSubjects there is a bit of redundancy that could be removed.

src/main/java/org/radarbase/management/web/rest/GroupResource.java
```diff
@@ -105,4 +113,62 @@ public GroupDTO getGroup(
...
+                    "Subject identification must be specified",
+                    GROUP, ERR_VALIDATION);
+            }
+            if (item.getId() != null && item.getLogin() != null) {
+                throw new BadRequestException(
+                    "Subject identification must be specify either ID or Login. " +
+                        "Do not provide both values to avoid potential confusion.",
+                    GROUP, ERR_VALIDATION);
+            }
+        }
```
blootsvoets, line 169: This could just be a function in the service and combine this code with the filtering code. Then no additional `allItems` list is needed, you just call
```java
parseSubjectSpecifications(subjectsToAdd, idsToAddList, loginsToAddList);
parseSubjectSpecifications(subjectsToRemove, idsToRemoveList, loginsToRemoveList);
```
in the service, with `idsToAddList`, etc. are empty array lists.

blootsvoets: If you still want to do it with streams, you could partition using collect with a map
```
partitioning = subjectsToAdd.stream()
    .collect(Collectors.groupingBy(s -> {
        Long id = s.getId();
        String login = s.getLogin();
        if (login != null && id != null) {
          return "duplicate";
        } else if (login != null) {
          return "login";
        } else if (id != null) {
          return "id";
        } else {
          return "empty";
        }
    });
if (partitioning.get("empty") != null) // bad request
if (partitioning.get("duplicate") != null) // bad request
partitioning.get("login").stream().map(::getLogin).collect(toList());
```
but it looks much more complicated this way.

aykl: Parsing, data fetching and validation combined in `getSubjectEntities`. Now `updateGroupSubjects` looks much cleaner.

blootsvoets (approved):
LGTM

blootsvoets:
Update - could you please create an additional test in `GroupResourceIntTest` and fix any issues with `./gradlew check`?

PR #630, opened 2021-11-10
Feature/organizations

aykl:
Implements the concept of organization in the Management Portal. This will allow managing multiple projects within the organization and delegate organization and project related tasks to the organization specific roles. 

#### Checklist:


blootsvoets:
FYI, I rebased the branch to let the commits make sense. Continue coding with
```
git pull --rebase
```

src/main/java/org/radarbase/management/service/dto/ProjectDTO.java
```diff
@@ -37,6 +37,8 @@
 
     private String organization;
 
+    private String organizationName;
+
```
blootsvoets, line 41: I'd expect an `OrganisationDTO` object here as well.

src/main/java/org/radarbase/management/domain/Organization.java
```diff
@@ -0,0 +1,114 @@
...
+    private String name;
+
+    @NotNull
+    @Column(name = "description", nullable = false)
+    private String description;
+
+    @NotNull
+    @Column(name = "location", nullable = false)
+    private String location;
+
```
blootsvoets, line 51: I'd expect a `@OneToMany` annotation to projects here.

blootsvoets (approved):
Looks good to me. A complete link between orgs and projects is not there yet, but it can be made at a later time.

src/main/java/org/radarbase/management/service/dto/ProjectDTO.java
aykl: I decided not to add it — we still have the old `organization` field. It would be great to remove it first once we are certain that nobody uses it — and re-use the field name. However, even this improvement is likely unnecessary, as any client that needs organization data can fetch it.

PR #631, opened 2021-11-15
Feature/delete group confirmation

aykl:
Description: Return 409 on DELETE {group} if the group has any subjects linked to it. Unlink subjects on deletion if the query parameter is set to true.

#### Checklist:


blootsvoets:
I thought that because of the foreign key and cascade settings, the groups were automatically removed from the subjects (no explicit query needed).

aykl:
You are correct, `unlinkAllFromGroup` is redundant. My mistake, it was faster to add the explicit query than to check whether they are unlinked implicitly, the focus was on adding the "does this group have any subjects?" check.

blootsvoets (approved):
LGTM

PR #634, opened 2021-11-17
Reactive filtering of subject component

blootsvoets:
Description:
- Make subject filtering reactive
- Refactor SubjectComponent to delegate some tasks to subcomponents
- Include filters in URL.

### DIFF
--- a/src/main/webapp/app/shared/util/reactive-filter.ts
+++ b/src/main/webapp/app/shared/util/reactive-filter.ts
@@ -17,14 +17,13 @@ import {
 } from "@ng-bootstrap/ng-bootstrap";
 
 export class ReactiveFilter<T> {
-  private readonly _value$: BehaviorSubject<T>;
-  private readonly _error$: BehaviorSubject<string>;
-  private readonly trigger$: Subject<T>;
+  protected readonly _value$: BehaviorSubject<T>;
+  protected readonly _error$: BehaviorSubject<string>;
+  protected readonly trigger$: Subject<T>;
 
   error$: Observable<string>
   rawValue$: Observable<T | null>
   value$: Observable<T | null>;
-  formattedValue$: Observable<string>;
 
   constructor(
     options?: ReactiveFilterOptions<T>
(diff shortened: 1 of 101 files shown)


src/main/webapp/app/shared/util/reactive-filter.ts
```diff
@@ -0,0 +1,141 @@
...
+    this._value$.next(null);
+    this._error$.next('');
+    this.trigger$.next(null);
+  }
+
+  complete() {
+    this._value$.complete();
+    this._error$.complete();
+    this.trigger$.complete();
+  }
```
peyman-mohtashami, line 77: This method is not used. Shouldn't we call it the onDestroy method of subject.component.ts?

peyman-mohtashami (commented):
LGTM
And, same as before, great effort and nice refactoring. Thank you.

src/main/webapp/i18n/nl/subject.json
```diff
@@ -26,7 +26,7 @@
                 "advancedFilter": "Uitgebreid filter",
                 "error": {
                     "invalidDate": "De datum is niet geldig.",
-                    "invalidRange": "De datum is niet geldig of ligt na de Van-datum."
+                    "invalidRange": "De periode is is niet geldig: Vanaf datum is na de Tot en met datum."
```
aykl, line 29:
```suggestion
                    "invalidRange": "De periode is niet geldig: Vanaf datum is na de Tot en met datum."
```

src/main/webapp/app/shared/util/reactive-filter.ts
blootsvoets: My editor says the same, but it is called in SubjectComponent line 162:
```
        for (let filtersKey in this.filters) {
             this.filters[filtersKey].complete();
         }
```

peyman-mohtashami: Yes, it's my mistake.

PR #636, opened 2021-11-18
Group removal warning

peyman-mohtashami:
Description: Removal of a group from project gives warning dialog

It provides a warning if a used group is being removed from a project. 
The backend will first return an HTTP 409 error → this triggers a warning “group has subjects, still want to delete”. 
If confirmed, it removes the group from the subjects who were assigned to this group with additional “unlinkSubjects=true” query param.


blootsvoets (approved):
LGTM

PR #642, opened 2021-12-07
Fix/permission updates

blootsvoets:
Allow updating roles by non sys-admin users. Use new API to enforce this.

### DIFF
--- a/src/main/java/org/radarbase/management/service/UserService.java
+++ b/src/main/java/org/radarbase/management/service/UserService.java
@@ -2,6 +2,8 @@ package org.radarbase.management.service;
 
 import org.radarbase.auth.authorization.RoleAuthority;
 import org.radarbase.auth.config.Constants;
+import org.radarbase.auth.exception.NotAuthorizedException;
+import org.radarbase.auth.token.RadarToken;
 import org.radarbase.management.config.ManagementPortalProperties;
 import org.radarbase.management.domain.Role;
 import org.radarbase.management.domain.User;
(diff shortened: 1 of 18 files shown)


src/main/java/org/radarbase/management/web/rest/UserResource.java
```diff
@@ -257,4 +251,36 @@ public ResponseEntity createUser(@RequestBody ManagedUserVM managedUserVm)
...
+    public ResponseEntity<Set<RoleDTO>> getUserRoles(@PathVariable String login)
+            throws NotAuthorizedException {
+        log.debug("REST request to read User roles: {}", login);
+        checkPermission(token, ROLE_READ);
+        return ResponseUtil.wrapOrNotFound(userService.getUserWithAuthoritiesByLogin(login)
+                        .map(UserDTO::getRoles));
+    }
+
+    /**
+     * PUT /users/:login/roles : get the "login" User roles.
```
aykl, line 272:
```suggestion
     * PUT /users/:login/roles : update the "login" User roles.
```

src/main/java/org/radarbase/management/service/UserService.java
```diff
@@ -189,6 +204,45 @@ public User createUser(UserDTO userDto) {
...
+                if (role.getProject().getOrganization() != null) {
+                    checkPermissionOnOrganizationAndProject(token, ROLE_UPDATE,
+                            role.getProject().getOrganization().getName(),
+                            role.getProject().getProjectName());
+                } else {
+                    checkPermissionOnProject(token, ROLE_UPDATE,
+                            role.getProject().getProjectName());
+                }
+            }
+            default -> { }
```
aykl, line 244: If we get an unexpected scope here, doesn't it indicate an exception-worthy bug?

aykl (approved):
LGTM.

src/main/java/org/radarbase/management/service/UserService.java
blootsvoets: Yes... In fact I only added the default branch to silence a PMD warning -- it's already an exhaustive switch for all enum possibilities. I'll remove it and then it will be a compilation error instead if it's not exhaustive.

aykl: Okay, if this is intentional I'm fine with it.

( However, a need to suppress a warning/ignore a newly introduced rule always leads to an obvious question :) )

PR #649, opened 2021-12-21
Feature/organization UI permissions enhancements

blootsvoets:
Feature/organization UI permissions enhancements, mainly, using scope in authorities response.


peyman-mohtashami (commented):
LGTM
I added the missed users to the CSV. Without them, I received the users without roles in the User Management list.

PR #651, opened 2022-01-11
Fix route guards

peyman-mohtashami:
- Fix route guards
- Fix frontend e2e test


src/main/webapp/app/entities/organization/organization-detail.component.html
```diff
@@ -4,7 +4,9 @@ <h2>
             <span [translate]="'managementPortalApp.organization.detail.title'"></span>
             {{organization.name}}
         </span>
-        <button type="button"
+        <button *ngIf="(principal.account$ | async).authorities.includes('ROLE_SYS_ADMIN') ||
+                    (principal.account$ | async).authorities.includes('ROLE_ORGANIZATION_ADMIN')"
+                type="button"
```
blootsvoets, line 9: If possible, add the `*ngIf="principal.account$ | async as account"` in the component above and use `account` in the button. This way just one subscription is created. If necessary, you can create a `<ng-container>` parent tag to put the `ngIf` in.

PR #657, opened 2022-01-28
Release 0.8.1

peyman-mohtashami:
Changes since version 0.8.0:

- Reduce dependency on JHipster
- Migrate to latest Angular version
- Add `participant_group` key to the subject's attribute.


nivemaham (approved):
LGTM.

PR #696, opened 2023-02-01
Continue pulling public key endpoints even if one fails

mpgxvii:
Currently, when a public key fails to be pulled from the public key endpoint, it throws an error and stops the rest of the public keys to be pulled (in the list of public key endpoints). This change will allow a warning to be shown instead of an error being thrown, which will continue the process of pulling the rest of the public keys.

### DIFF
--- a/radar-auth/src/main/java/org/radarbase/auth/authentication/TokenValidator.java
+++ b/radar-auth/src/main/java/org/radarbase/auth/authentication/TokenValidator.java
@@ -246,7 +246,8 @@ public class TokenValidator {
                 return algorithmLoader.loadAlgorithmsFromJavaWebKeys(publicKeyInfo);
             } else {
                 // Log and Continue Pulling next Endpoints, if any
-                LOGGER.warn("Invalid token signature. Could not load newer public keys");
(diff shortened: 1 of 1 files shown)


radar-auth/src/main/java/org/radarbase/auth/authentication/TokenValidator.java
```diff
@@ -246,7 +246,8 @@ private List<Algorithm> algorithmFromServerPublicKeyEndpoint(URI serverUri) thro
                 return algorithmLoader.loadAlgorithmsFromJavaWebKeys(publicKeyInfo);
             } else {
                 // Log and Continue Pulling next Endpoints, if any
-                LOGGER.warn("Invalid token signature. Could not load newer public keys");
+                LOGGER.warn("Invalid token signature. Could not load " 
+                        + "public key from {}", serverUri.toURL());
```
blootsvoets, line 250:
```suggestion
                LOGGER.warn("Could not load newer public keys from {}", serverUri.toURL());
```

blootsvoets (approved):
LGTM

PR #702, opened 2023-02-23
Fix Docker build failing

mpgxvii:
Description: 
- Update Docker build to just build the stage 1 image on linux/amd64


blootsvoets (approved):

mpgxvii:
> I've changed the PR to just build the stage 1 image on linux/amd64, solving all issues. In MP 2.0 the build works for both platforms. I've also changed the Docker base images to temurin, because the alpine images are not available for linux/arm64.

Okay, I see. I wasn't sure if I could change the platform, in case it will affect other things downstream. Thanks so much!

PR #704, opened 2023-03-13
Fix pair token dialog with delete

blootsvoets:
- fixes autowiring of RadarToken
- fixes JS request handling for delete pair request - use HttpResponse instead of contents of empty 204 response.


nivemaham (approved):
LGTM

PR #715, opened 2023-09-11
709 make managementportal user properties configurable (backend)

### LINKED ISSUE #709: Make ManagementPortal user properties configurable
Specifically the following fields should be made optional as specified in frontend setup:
- Name
- Birth date
- Group

This is just a frontend customisation, those fields are optional in the backend already. Specifically, the following pages should be modified:
- Participant edit / info screen
- Per-property filtering turned off or on
- Group tab hidden / shown

The frontend should get informed of what features to enable via a REST request. Ideally, this request goes to ManagementPortal backend, otherwise the Dockerfile needs to add an entrypoint script to fill in the configured values when the container is started up.
```[tasklist]
### Tasks
- [x] add all to group checkboxes should be hidden
- [x] remove hidden fields in pairing dialogue
- [x] handle info-screen
- [x] handle participant-edit
- [x] handle per-property filtering
- [x] handle group tab
- [x] implement requirements on the backend
```

Bdegraaf1234:
Description: Add a SiteSettings configuration to managementPortal to store settings which can be accessed prior to user Authentication. The first setting will be which subject fields to hide.

Fixes #709

#### Checklist:


Bdegraaf1234:
I did not update the readme, I don't think its needed?

src/main/java/org/radarbase/management/service/SiteSettingsService.java
```diff
@@ -0,0 +1,50 @@
...
+     * Convert a {@link SiteSettingsDto} to a {@link SiteSettings} object.
+     *
+     * @param siteSettingsDto The DTO object to convert
+     * @return the newly created object
+     */
+    public SiteSettings createSiteSettings(SiteSettingsDto siteSettingsDto) {
+        SiteSettings siteSettings = new SiteSettings();
+        siteSettings.setHiddenSubjectFields(siteSettingsDto.getHiddenSubjectFields());
+        return siteSettings;
+    }
```
blootsvoets, line 44: This is not used, right? It can be removed

src/main/java/org/radarbase/management/service/SiteSettingsService.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Convert a {@link SiteSettings} to a {@link SiteSettingsDto} object.
+     * @param siteSettings The object to convert
+     * @return the newly created DTO object
+     */
+    public SiteSettingsDto createSiteSettingsDto(SiteSettings siteSettings) {
+
+        SiteSettingsDto siteSettingsDto;
+        siteSettingsDto = new SiteSettingsDto();
```
blootsvoets, line 28: Usually, this is a single statement
```suggestion
        SiteSettingsDto siteSettingsDto = new SiteSettingsDto();
```

src/main/resources/config/application.yml
```diff
@@ -114,6 +114,8 @@ managementportal:
         keyStorePassword: radarbase
         signingKeyAlias: radarbase-managementportal-ec
         enablePublicKeyVerifiers: false
+    siteSettings:
+      hiddenSubjectFields: [person_name, date_of_birth, group]
```
blootsvoets, line 118: By default, this can be an empty list, since we don't want to hide these features by by default.

blootsvoets (commented):
Alright, looks good. I've added some small comments.

Bdegraaf1234:
closing, moved to #716

PR #716, opened 2023-09-13
709 make managementportal user properties configurable

### LINKED ISSUE #709: Make ManagementPortal user properties configurable

Bdegraaf1234:
Description: Frontend component of #715. We retrieve the SiteSettings through the API and use it to configure which fields should be hidden from several subject views.

Fixes #709

#### Checklist:


src/main/webapp/app/shared/subject/sitesettings.service.ts
```diff
@@ -0,0 +1,18 @@
...
+export class SiteSettingsService {
+
+    private siteSettingsResourceUrl = 'api/sitesettings';
+
+
+    constructor(private http: HttpClient) {
+    }
+
+    getSiteSettings(): Observable<SiteSettings> {
+        return this.http.get<SiteSettings>(`${this.siteSettingsResourceUrl}`);
```
peyman-mohtashami, line 16: The `siteSettingsResourceUrl` is a string, so you can just put it as arguments without template string (`)
```
return this.http.get<SiteSettings>(this.siteSettingsResourceUrl);
```

src/main/webapp/app/shared/subject/subject-detail.component.ts
```diff
@@ -16,10 +17,12 @@ export class SubjectDetailComponent implements OnInit, OnDestroy {
     subject: Subject;
     private subscription: any;
     private eventSubscriber: Subscription;
+    public siteSettings$: Observable<SiteSettings>;
```
peyman-mohtashami, line 20: You can merge this line and line #25 here.
```
public siteSettings$ = this.siteSettingsService.getSiteSettings();
```
And you can remove the Observable import line on top.

src/main/webapp/app/shared/subject/subject-pair-dialog.component.ts
```diff
@@ -28,12 +29,14 @@ export class SubjectPairDialogComponent implements OnInit, OnDestroy {
     pairInfo: PairInfo = null;
     selectedClient: OAuthClient = null;
     allowPersistentToken = false;
+    public siteSettings$: Observable<SiteSettings>;
```
peyman-mohtashami, line 32: Same as the above comment:
You can merge this line and line 51 here.
```
public siteSettings$ = this.siteSettingsService.getSiteSettings();
```
And you can remove the Observable import line on top and `public` keyword.

src/main/webapp/app/shared/subject/subject-pair-dialog.component.html
```diff
@@ -26,15 +26,15 @@
                        [(ngModel)]="subject.externalId" [disabled]="subject.id" >
             </div>
 
-            <div *ngIf="subject.personName" >
+            <div *ngIf="!(siteSettings$ | async).hiddenSubjectFields.includes('person_name') && subject.personName" >
```
peyman-mohtashami, line 29: Did you check for a null state? e.g. if siteSettings is null to undefined?

peyman-mohtashami (commented):
Is there a modification for the group tab, and subjects filter?

src/main/webapp/app/shared/subject/subject-pair-dialog.component.html
Bdegraaf1234: as discussed f2f: sitesettings is initialized as an empty list during startup. If the get request goes wrong it is because the backend is failing as a whole which will be caught elsewhere, so imo no explicit error-handling is needed here.

src/main/webapp/app/shared/subject/subject-detail.component.html
```diff
@@ -39,11 +42,14 @@ <h2><span [translate]="'managementPortalApp.subject.detail.title'"></span>: {{su
...
-        <dt><span [translate]="'managementPortalApp.subject.dateOfBirth'"></span></dt>
-        <dd>{{subject.dateOfBirth | date:'yyyy-MM-dd' }}</dd>
+        <div *ngIf="!(siteSettings$ | async).hiddenSubjectFields.includes('person_name')">
+            <dt><span [translate]="'managementPortalApp.subject.personName'"></span></dt>
+            <dd>{{subject.personName}}"</dd>
+        </div>
+        <div *ngIf="!(siteSettings$ | async).hiddenSubjectFields.includes('date_of_birth')">
+            <dt><span [translate]="'managementPortalApp.subject.dateOfBirth'"></span></dt>
+            <dd>{{subject.dateOfBirth | date:'yyyy-MM-dd' }}</dd>
+        </div>
```
blootsvoets, line 52: To avoid a lot of subscriptions from a component, you can create a single subscription in a parent div and store and reuse the result:
```html
<div *ngIf="(siteSettings$ | async).hiddenSubjectFields as hiddenFields">
   ...
   <div *ngIf="!hiddenFields.includes('person_name')"...>...</div>
   <div *ngIf="!hiddenFields.includes('date_of_birth')"...>...</div
</div>
```
strangely this assignment only works in an ngIf, so the value must evaluate to true. Fortunately, objects and arrays always evaluate to true.

Bdegraaf1234:
closing #715, I pulled the changes in here, things were getting mixed anyway we may as well review them all in a bundled manner.

src/main/webapp/app/shared/subject/subject-pair-dialog.component.html
```diff
@@ -10,39 +10,39 @@
...
     <div class="modal-body">
         <div class="form-group">
-            <div>
-                <label for="login" class="form-control-label"
-                       [translate]="'managementPortalApp.subject.login'">
-                </label>
-                <input id="login" type="text" class="form-control" name="login" #loginInput="ngModel"
-                       [(ngModel)]="subject.login" [disabled]="subject.id" >
-            </div>
+            <div *ngIf="(siteSettings$ | async).hiddenSubjectFields as hiddenFields">
```
blootsvoets, line 13: Instead of adding an additional div layer, you can add this `ngIf` to the `<div class="form-group">` instead. Or only add this additional div layer to the properties that are actually modified by the siteSettings.

blootsvoets (commented):
Looks good to me. I have a few questions but no blockers.

blootsvoets (approved):

PR #718, opened 2023-09-15
707 add print button to pair dialog

### LINKED ISSUE #707: Add print button to pair dialog
Allow the pair QR code dialog to be printed out without any additional data, e.g. using a print icon. Right now this will also print all backgrounds and subject lists.

Bdegraaf1234:
Description: Adding a print button to the subject-pair dialogue.

Fixes #707

#### Checklist:

### DIFF
--- a/src/main/webapp/app/shared/subject/subject-pair-dialog.component.ts
+++ b/src/main/webapp/app/shared/subject/subject-pair-dialog.component.ts
@@ -1,4 +1,4 @@
-import { Component, Inject, OnDestroy, OnInit } from '@angular/core';
+import { Component, Inject, OnDestroy, OnInit} from '@angular/core';
 import { DatePipe, DOCUMENT } from '@angular/common';
 import { HttpResponse } from '@angular/common/http';
 import { ActivatedRoute, Params } from '@angular/router';
--- a/src/main/webapp/app/shared/subject/subject-pair-dialog.component.ts
+++ b/src/main/webapp/app/shared/subject/subject-pair-dialog.component.ts
@@ -14,11 +14,13 @@ import { SubjectPopupService } from './subject-popup.service';
 import { Subject } from './subject.model';
 import { ObservablePopupComponent } from '../util/observable-popup.component';
 import { map, switchMap, tap } from 'rxjs/operators';
+import { JhiMainComponent } from '../../layouts/main/main.component';
+import { PrintService } from '../util/print.service';
 
 @Component({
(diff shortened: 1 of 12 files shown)


src/main/webapp/app/shared/subject/subject-pair-dialog.component.ts
```diff
@@ -35,9 +37,23 @@ export class SubjectPairDialogComponent implements OnInit, OnDestroy {
...
     }
+    async exportHtmlToPDF() {
+        // let pp: PrintService = this.printService;
+        // window.onafterprint = function(){
+        //     console.log("Printing completed...");
+        //     //pp.togglePrintLock().then(_ => {})
+        // }
+
+        this.printService.setPrintLockTo(true);
+        await new Promise(r => setTimeout(r, 0.000000000000000000000000000000000000000000000000000001));
```
Bdegraaf1234, line 52: I think the implementation is fine except for the need for this delay and I cannot figure out how to remove it...

blootsvoets: It’s needed because otherwise angular cannot do change detection while you’re in a blocking function. Alternatively, you lock the print lock when the modal opens and release when the modal disappears. That way if they use the browser to print, they still get the same view. And we can expect that when users print a modal, they are not interested in the background.

Bdegraaf1234: Thanks, that fixed it indeed!

blootsvoets (approved):
The code looks good to me. Does the print view look as expected?

Bdegraaf1234:
final printview: 
image

src/main/webapp/app/shared/util/print.service.ts
```diff
@@ -4,14 +4,11 @@ import { BehaviorSubject } from "rxjs";
...
-
-    constructor(
-    ) {
-        this.isPrintLocked$.subscribe();
-    }
+    isPrintLocked$:  BehaviorSubject<boolean> = new BehaviorSubject<boolean>(false);
 
     setPrintLockTo(setTo: boolean) {
-        this.isPrintLocked$.next(setTo)
+        setTimeout(() => {
```
blootsvoets, line 10: this timeout should not be necessary, is it?

Bdegraaf1234: You are right, it is not.

thought I checked that thoroughly, my bad.

blootsvoets (approved):

PR #723, opened 2023-09-22
722 wait no longer needed in cypress tests

### LINKED ISSUE #722: Wait no longer needed in cypress tests
We have a number of flaky cypress tests, for which we use the cy wait function.

However, it seems that after #716 added retries, this should not be necessary anymore.

We should remove the calls to cy.wait where possible.

Bdegraaf1234:
Description: Remove unneeded wait commands from cypress tests

Fixes #722 

#### Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -75,7 +75,7 @@ jobs:
           yarn e2e
           ./gradlew --stop
 
-      - name: cypress-screenshots-failed-tests
+      - name: Upload screenshots of failed e2e tests
         if: always()
         uses: actions/upload-artifact@v3
         with:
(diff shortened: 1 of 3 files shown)


Bdegraaf1234:
Unfortunately, it seems that the need for wait is not removed.

One way to achieve non-arbitrary wait times would be to add guards by intercepting the xhr requests.

```
cy.intercept("GET", "/api/projects/radar/subjects?size=20&sort=login,desc").as("sort")
cy.get('jhi-subjects #order-by-desc').click();
cy.wait("@sort");
```
However, this seems like quite a bit more work and the wait times are not that long. I have also failed to find the logic in where the wait commands are needed.

I was able to remove three out of 4 without causing problems here.

blootsvoets:
Failed tests with a lot of try / retry probably will still need the original wait. Once fixed, can this be merged or is it still WIP?

Bdegraaf1234:
Turned out to be pretty small changes, I left the alternative wait strategy in with a comment. ready for review

blootsvoets (approved):
LGTM

PR #725, opened 2023-09-27
724 redirect traffic from to managementportal

Bdegraaf1234:
Description: Change some dev-settings to simplify development by making web-requests sent to the Angular Live Development Server (ALDS) follow the same routes as web-requests sent to (locally hosted) prod builds (i.e. the docker images used in this and other repos)

One "breaking" change is that the ALDS is now hosted at port 8081, as this is the default port we host management portal on when hosted in conjunction with other services.

furthermore, requests pointing to {ALDS}/managementportal are redirected to root, and a description of the client credential authorization workflow was added to the readme

Fixes #724

#### Checklist:
- [ ] The [Main workflow](https://github.com/RADAR-base/ManagementPortal/actions/workflows/main.yml) has succeeded
- [ ] The [Gatling tests](https://github.com/RADAR-base/ManagementPortal#other-tests) have passed
- [ ] I have logged into the portal running locally with default admin credentials
- [ ] I have updated the README files if this change requires documentation update
- [ ] I have commented my code, particularly in hard-to-understand areas

Issue #717, opened 2023-09-14
Test production mode in integration test


blootsvoets:
Currently, we only test ManagementPortal in dev mode in Github actions. This does not test the PostgreSQL integration. Ideally, we add a step to the Github Actions docker job to run the created docker image with docker-compose and test if it functions. A minimal implementation tests if ManagementPortal becomes healthy.

An example would be the first two services in https://github.com/RADAR-base/radar-app-config/blob/master/docker-compose.yml.

Bdegraaf1234:
When do we want to run prod-tests? 

Since currently the prod e2e-tests and dev e2e-tests are identical it seems excessive to run both as part of the `main.yml` workflow. which do we run in main?

For `release.yml` and `pre-release.yml` I suggest we just run both.

Alternatively, we could probably make a profile of which tests to run in prod/dev but I don't think that would be worth it at this point.

Bdegraaf1234:
They run in pre-release now :)

PR #733, opened 2023-10-04
fix/append-managementportal-to-baseurl

### LINKED ISSUE #732: append /managementportal to baseurl
The issue described in #724 persists despite the fix in #725, as we decided to reroute traffic from `<base_url>/managementportal` to `<base_url>/`. 

We will now alter the managementportal so that `/managementportal` is included in the baseurl to prevent excessive rerouting, which only complicates manners.

current problems:
- login in dev setup only works when the proxy.conf is used, thus requiring `yarn start`
- issue is with the pair request on local development: it contains a duplicate `/` in the meta token URL.

Bdegraaf1234:
No more rerouting in the Angular live development server (ALDS). server.servlet.context-path and managementportal.common.managementportalbaseurl changed in dev to match prod.

Description: change application-dev.properties to more closely match application-prod.properties and revisit some changes made in #725 

Fixes #732

#### Checklist:

### DIFF
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:11-jre-slim
+FROM eclipse-temurin:17-jdk
 
 ENV SPRING_OUTPUT_ANSI_ENABLED=ALWAYS \
     JHIPSTER_SLEEP=0
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM eclipse-temurin:17-jdk
+FROM eclipse-temurin:17-jre
 
 ENV SPRING_OUTPUT_ANSI_ENABLED=ALWAYS \
     JHIPSTER_SLEEP=0
--- a/src/main/resources/config/application-dev.yml
+++ b/src/main/resources/config/application-dev.yml
@@ -78,8 +78,9 @@ spring:
 server:
     port: 8080
     servlet:
-        session.cookie.secure: false
-        context-path: /managementportal
+      session.cookie.secure: false
+      context-path: /managementportal
+
 
 
 # ===================================================================
--- a/src/main/resources/config/application-dev.yml
+++ b/src/main/resources/config/application-dev.yml
@@ -80,6 +80,9 @@ server:
     servlet:
       session.cookie.secure: false
       context-path: /managementportal
+      session:
+        cookie:
+          path: /
 
 
 
--- a/src/main/resources/config/application-dev.yml
+++ b/src/main/resources/config/application-dev.yml
@@ -79,6 +79,7 @@ server:
     port: 8080
     servlet:
         session.cookie.secure: false
+        context-path: /managementportal
 
 
 # ===================================================================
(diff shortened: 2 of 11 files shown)


Bdegraaf1234:
this is a working setup but still needs the ALDS to handle the /magementportal/oauth/* requests.

did not look at the double `/` issue yet

blootsvoets (commented):
Also this would need the new base href value: https://github.com/RADAR-base/ManagementPortal/blob/67403f8e93f2570b754cd1b7e027ea3b15276e3d/src/main/webapp/index.html#L4

src/main/resources/config/application-dev.yml
```diff
@@ -87,8 +88,8 @@ server:
 # ===================================================================
 managementportal:
     common:
-        baseUrl: http://localhost:9000 # Modify according to your server's URL
-        managementPortalBaseUrl: http://localhost:8080/
+        baseUrl: http://localhost:8081 # Modify according to your server's URL
```
Bdegraaf1234, line 91: changes here do not seem to affect anything, but it was a mismatch between the angular dev server port after changes in #725 so I set it to match again

src/main/resources/config/application-dev.yml
```diff
@@ -87,8 +88,8 @@ server:
...
+        managementPortalBaseUrl: http://localhost:8080/managementportal
```
Bdegraaf1234, line 92: The lack of trailing / should fix the broken meta-token url

blootsvoets (approved):

blootsvoets (changes requested):
I see that the gatling tests are still referring to the old URL. Please update those as well. See `src/gatling/scala/ManagementPortalSimulation`

src/main/resources/config/application-dev.yml
Bdegraaf1234: indeed, just used for building mail and metatokens

Bdegraaf1234:
It seems like the current production builds are not compatible with an ALDS proxy/frontend due to the addition of the `/managementportal/` context-path.

There is an authentication problem when using accessing via a front-end (cypress, ALDS), the token is successfully retrieved but not attached to the subsequent requests (specifically, `/api/organizations` and `/api/projects`).

The problem is over my head at this point, but I think it is related to #717 as it prevents us from running cypress tests on a production environment. I cannot find any place where this [script](https://github.com/RADAR-base/ManagementPortal/blob/master/src/test/bash/run-prod-e2e.sh) which should achieve this is ran.

blootsvoets:
Perhaps two proxy config are needed then — one with path rewrite and one without?

Bdegraaf1234:
I think the proxy stuff was a bit of red herring, it really was a problem with the cookie not propagating somehow.

I now manually set the path of the cookie to `/` (mirroring the situation we have in the k3s-test deployment). This ensures that all requests belong to the same session, and the authentication works again.

I think we should now be able to run the e2e tests on a production build and remove quite a few distinct commands between dev and prod builds.

Bdegraaf1234:
Ready for review.

Ran the e2e tests on a local production build, 2 failing tests. Will make a branch for #717 to pick this up next week.

src/test/bash/run-prod-e2e.sh
```diff
@@ -27,7 +27,7 @@ set +e
 # wait for app to be up
 yarn run wait-for-managementportal-prod
```
blootsvoets, line 28: Should be `wait-for-managementportal`

src/main/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openjdk:11-jre-slim
+FROM eclipse-temurin:17-jdk
```
blootsvoets, line 1: Since nothing is built in the docker file, this can be `eclipse-temurin:17-jre`

blootsvoets (commented):
Looks good, nice that you got it fixed! Some small notes.

blootsvoets (approved):
LGTM

PR #744, opened 2023-11-10
Downstream testing trial

Bdegraaf1234:
Description: Added a github action which, on merging a branch into dev/master triggers downstream tests agains the built docker image

An alternative way to do this would be to pull the downstream repo and run the tests, but this seems more messy. It does however keep any failing pipelines close to the source (in this case, the changes to management portal)

#### Checklist:


Bdegraaf1234:
I think we can merge this, the largest change are the docker-compose files but a restructuring here is probably needed for the ory change because this will add a number of components that need to be launched to run MP

baixiac (commented):
LGTM and one comment added.

Also, I know this falls outside this PR's scope but probably worth noting: Postgres 11 has reached its EOL.

Bdegraaf1234:
The most important changes here are the ones in main.yml, the rest is probably redundant when you update the base

PR #829, opened 2024-05-13
Remove unused hikari config from application properties

mpgxvii:
- Getting an error: `java.lang.IllegalStateException: The configuration of the pool is sealed once started. Use HikariConfigMXBean for runtime changes.`
- I think these were not used and tested for a while (in previous releases) and was used in the latest release (which resulted in the error), but if this is needed, we would need a custom implementation of the `HikariConfig`

### DIFF
--- a/src/main/resources/config/application-prod.yml
+++ b/src/main/resources/config/application-prod.yml
@@ -25,12 +25,6 @@ spring:
     name: managementportal
     username: radarbase
     password: radarbase
-    hikari:
-      data-source-properties:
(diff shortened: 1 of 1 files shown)


yatharthranjan (approved):
LGTM, thanks.

Bdegraaf1234 (approved):
Looks good to me!

I recall encountering this problem and not being able to fix it with just these changes, but I think that was in the backport branch (v0.8). Just something to keep an eye out for.

Merging

PR #837, opened 2024-05-22
Update Docker build

mpgxvii:
Description: 

- Updates Docker build: allow front-end env vars to be updated from `docker-compose` file
- Make Kratos redirect url the origin url (in this case MP url)

Tested by running:
```
docker-compose -f "src/main/docker/managementportal_local/docker-compose.yml" up -d --build 
```

Then open: `http://127.0.0.1:8080/managementportal`


Bdegraaf1234:
Thanks @mpgxvii this looks great :)

We had problems with the deployed containers not picking up the variables in the helm charts. Am I correct in assuming this will likely fix that as well?

Merging.

Bdegraaf1234 (approved):

PR #848, opened 2024-06-04
Disable broken UI tests

### LINKED ISSUE #799: E2E tests fail due to changed UI
Ory-based ui tests should be added. There are good examples in the ory repository:

https://github.com/ory/kratos/blob/fc2cecfbeab811aa1a851f953b6bc2a4c119c412/test/e2e/cypress/integration/profiles/mfa/totp.spec.ts

Bdegraaf1234:
Description: currently the e2e tests are broken due to the addition of ory-based identity management. While less than ideal this is preventing the CI to run and can thus introduce unnecessary errors. For the time being we will disable the e2e tests that require logging in until the new auth flows are mature enough that they can be consistently tested.

Partially fixes #799

#### Checklist:

### DIFF
--- a/src/main/docker/non_managementportal/docker-compose.yml
+++ b/src/main/docker/non_managementportal/docker-compose.yml
@@ -21,8 +21,6 @@ services:
     networks:
       - db
       - default
-    ports:
-      - "5432:5432"
 
   kratos-selfservice-ui-node:
     extends:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -69,12 +69,13 @@ jobs:
       - name: End to end tests
         id: e2e
         run: |
-          cp src/test/resources/config/keystore.p12 src/main/resources/config/keystore.p12
-          ./gradlew bootRun &>mp.log </dev/null &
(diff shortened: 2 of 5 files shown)


Bdegraaf1234:
Ready to merge 👍

src/main/docker/non_managementportal/docker-compose.yml
```diff
@@ -21,8 +21,6 @@ services:
     networks:
       - db
       - default
-    ports:
-      - "5432:5432"
```
yatharthranjan, line 25: probably needed if need to access from outside docker (as the top comment suggests)?

Bdegraaf1234: It is, but this is also defined in the file: src/main/docker/postgres.yml

So it's a question of where do we define it. My docker complained that the port was already assigned, which is why I removed one of the definitions. Since this port is so standard  figured it could be in the postgres config rather than in the docker_compose.

Do you suggest moving it here?

yatharthranjan (approved):
LGTM

PR #879, opened 2024-07-03
Fetch kratos login url from backend

mpgxvii:
Description: 

- Fetch kratos login url from backend
- Adding the kratos URL to the Docker build only updates it during build time and not during runtime
- This can now be set from Docker/k8s with `MANAGEMENTPORTAL_IDENTITYSERVER_LOGINURL`


src/main/java/org/radarbase/management/web/rest/LoginEndpoint.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.web.bind.annotation.RestController
+import org.springframework.web.servlet.view.RedirectView
+
+@RestController
+class LoginEndpoint
+    @Autowired
+    constructor(
+        @Autowired private val managementPortalProperties: ManagementPortalProperties,
+    ) {
+        @GetMapping("/oauth/login")
```
yatharthranjan, line 16: should we name these `api/redirect/login` and `api/redirect/account`?

yatharthranjan (dismissed):
Nice one! Just a minor comment.

src/main/java/org/radarbase/management/web/rest/LoginEndpoint.kt
mpgxvii: Yes made the changes.

PR #882, opened 2024-07-05
Fix/login redirects

Bdegraaf1234:
Login redirects were not set correctly in #879, additionally the antmatchers were not set in all required locations.

This should resolve the issue and add some other small improvements.


Bdegraaf1234:
@mpgxvii please correct me if I am wrong but for me therequests to `/api/redirect/login/` returned 401 at some point. I am double checking this as we speak though

mpgxvii:
> @mpgxvii please correct me if I am wrong but for me therequests to `/api/redirect/login/` returned 401 at some point. I am double checking this as we speak though

The previous security config were working for me. It was just the front end url that didn't work. Let me double check as well.

Bdegraaf1234:
I think you were right pauline, apologies. I will remove the lines, can you review the this PR?

mpgxvii (approved):
LGTM, sorry for the confusion. Thanks so much!

PR #888, opened 2024-07-08
Add kratos subject webhook endpoint

mpgxvii:
Description: 

- Add kratos webhook endpoint to allow automatic creation of subjects after kratos registration
- After subject is created, this adds back session id information such as scope, roles, to kratos id

Solves https://github.com/RADAR-base/hydra-kratos/issues/43


src/main/java/org/radarbase/management/web/rest/KratosEndpoint.kt
```diff
@@ -0,0 +1,83 @@
...
+        @PostMapping("/subjects")
+        @Timed
+        @Throws(URISyntaxException::class, NotAuthorizedException::class)
+        fun createSubject(@RequestBody webhookDTO: KratosSubjectWebhookDTO): ResponseEntity<SubjectDTO> {
+            val projectName = webhookDTO.project_id
+            val projectDto = projectService.findOneByName(projectName!!) 
+            // TODO: Add permission
+            // authService.checkPermission(Permission.SUBJECT_CREATE, { e: EntityDetails -> e.project(projectName) })
+            val subjectDto = SubjectDTO()
+            subjectDto.externalId = webhookDTO.identity_id
```
mpgxvii, line 61: @yatharthranjan @Bdegraaf1234 I'm currently just mapping the Kratos `identity_id` to the MP `subject_id` directly. Should I add a `kratos_` prefix? Or is this fine?

src/main/java/org/radarbase/management/web/rest/KratosEndpoint.kt
```diff
@@ -0,0 +1,83 @@
...
+         *
+         * @param subjectDto the subjectDto to create
+         * @return the ResponseEntity with status 201 (Created) and with body the new subjectDto, or
+         * with status 400 (Bad Request) if the subject has already an ID
+         * @throws URISyntaxException if the Location URI syntax is incorrect
+         */
+        @PostMapping("/subjects")
+        @Timed
+        @Throws(URISyntaxException::class, NotAuthorizedException::class)
+        fun createSubject(@RequestBody webhookDTO: KratosSubjectWebhookDTO): ResponseEntity<SubjectDTO> {
```
yatharthranjan, line 55: is this endpoint authorised? we need it to be authorised

src/main/java/org/radarbase/management/web/rest/KratosEndpoint.kt
Bdegraaf1234: I would probably add a field, at least I did for `user`. for example, the external id is an easily editable field in many places.

can be a new PR/issue though

src/main/java/org/radarbase/management/web/rest/KratosEndpoint.kt
```diff
@@ -0,0 +1,83 @@
...
+                )
+            }
+            val result = subjectService.createSubject(subjectDto)
+            return ResponseEntity.created(ResourceUriService.getUri(subjectDto))
+                .headers(HeaderUtil.createEntityCreationAlert(EntityName.SUBJECT, result?.login))
+                .body(result)
+        }
+
+        companion object {
+            private val logger = LoggerFactory.getLogger(TokenKeyEndpoint::class.java)
```
Bdegraaf1234, line 81:
```suggestion
            private val logger = LoggerFactory.getLogger(KratosEndpoint::class.java)
```

Bdegraaf1234 (approved):
Small change in a suggestion but otherwise LGTM,

src/main/java/org/radarbase/management/web/rest/KratosEndpoint.kt
yatharthranjan: Ah, yes, that's a good point. We would not want a user to change it so that a new field could be nice. Alternatively, we can remove the modification of externalId from the UI

src/main/java/org/radarbase/management/web/rest/KratosEndpoint.kt
yatharthranjan: Can the token/session cookie not be sent through the body?

mpgxvii: There are only 2 types of Auth it accepts in the header `Basic` and `API`. So I think for now we can just send through the body?

Bdegraaf1234:
LGTM, ready to merge?

yatharthranjan:
Hi @Bdegraaf1234, we are waiting to merge till the auth is sorted (depends on jwt and claims being added)

Bdegraaf1234:
Alright, converted to draft then :)

src/main/java/org/radarbase/management/service/UserService.kt
```diff
@@ -166,20 +154,9 @@ class UserService @Autowired constructor(
         user.password = passwordService.generateEncodedPassword()
         user.resetKey = passwordService.generateResetKey()
         user.resetDate = ZonedDateTime.now()
-        user.activated = true
+        user.activated = false
```
Bdegraaf1234, line 157: When do we activate users?

src/main/java/org/radarbase/management/service/UserService.kt
```diff
@@ -365,28 +322,7 @@ class UserService @Autowired constructor(
...
-        user.email = email
-        log.debug("Set admin email to: {}", email)
-
-        // there is no identity for this user, so we create it and save it to the IDP
-        val id = identityService.saveAsIdentity(user)
-        // then save the identifier and update our database
-        user.identity = id?.id
-        return userMapper.userToUserDTO(user)
-            ?: throw Exception("Admin user could not be converted to DTO")
     }
```
Bdegraaf1234, line 326: Is there an alternative flow to get admin access? we should make sure to describe it, the current flow is in the readme I believe so it should be removed if not functional anymore

src/main/java/org/radarbase/management/service/UserService.kt
```diff
@@ -412,14 +348,13 @@ class UserService @Autowired constructor(
...
-    @Transactional(readOnly = true)
-    /**
-     * Get the current user.
-     * @return the currently authenticated user, or null if no user is currently authenticated
-     */
-    fun getUserWithAuthorities(): User? {
-        return SecurityUtils.currentUserLogin?.let { userRepository.findOneWithRolesByLogin(it) }
-    }
+    @get:Transactional(readOnly = true)
+    val userWithAuthorities: User?
```
Bdegraaf1234, line 352: If we are renaming this I would make it `currentUserWithAuthorities`

```suggestion
    val currentUserWithAuthorities: User?
```

Bdegraaf1234 (commented):
Some comments, If you tested I won't hold this back but there are some chores we should keep tabs on

src/main/java/org/radarbase/management/service/UserService.kt
mpgxvii: We activate them after they verify their email.

src/main/java/org/radarbase/management/service/UserService.kt
mpgxvii: After discussions, I'll just revert the changes.

mpgxvii:
Closing, moved here: #945

PR #904, opened 2024-07-24
Remove unused hikari config from application properties

mpgxvii:
Description: 
- Remove unused hikari config from application properties

Fixes:
- Error: `java.lang.IllegalStateException: The configuration of the pool is sealed once started. Use HikariConfigMXBean for runtime changes.`

### DIFF
--- a/src/main/resources/config/application-prod.yml
+++ b/src/main/resources/config/application-prod.yml
@@ -25,12 +25,6 @@ spring:
     name: managementportal
     username: radarbase
     password: radarbase
-    hikari:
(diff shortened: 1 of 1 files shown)


Bdegraaf1234:
I'm having a deja vu  here @mpgxvii, was the previous pr never merged? 

Also, why not merge into dev first?

If we want a release, we should keep in mind that there are pretty big breaking changes here right? no kratos=no mp

I would call it 3.0.0-alpha or something?

yatharthranjan (approved):
LGTM

Issue #915, opened 2024-08-01
Expose a new public endpoint for fetching projects


this-Aditya:
Create a new public endpoint for exposing the project's info. This should not expose all the fields of the ProjectDto, but instead a subset of it, which includes projectName, description, location, and projectStatus.

PR #923, opened 2024-08-05
Added public endpoint for fetching subset of projects info

this-Aditya:
Added a new endpoint for fetching the project data to be used for the self-enrollment portal

Fixes https://github.com/RADAR-base/ManagementPortal/issues/915

### DIFF
--- a/src/main/java/org/radarbase/management/config/OAuth2ServerConfiguration.kt
+++ b/src/main/java/org/radarbase/management/config/OAuth2ServerConfiguration.kt
@@ -129,7 +129,7 @@ class OAuth2ServerConfiguration(
                 .skipUrlPattern(HttpMethod.GET, "/management/health")
                 .skipUrlPattern(HttpMethod.POST, "/oauth/token")
                 .skipUrlPattern(HttpMethod.GET, "/api/meta-token/*")
-                .skipUrlPattern(HttpMethod.GET, "/api/lite-projects")
+                .skipUrlPattern(HttpMethod.GET, "/api/public/projects")
                 .skipUrlPattern(HttpMethod.GET, "/api/sitesettings")
                 .skipUrlPattern(HttpMethod.GET, "/api/redirect/**")
                 .skipUrlPattern(HttpMethod.GET, "/api/logout-url")
@@ -174,7 +174,7 @@ class OAuth2ServerConfiguration(
                 .hasAnyAuthority(RoleAuthority.SYS_ADMIN_AUTHORITY)
                 .antMatchers("/api/profile-info").permitAll()
                 .antMatchers("/api/sitesettings").permitAll()
-                .antMatchers("/api/lite-projects").permitAll()
+                .antMatchers("/api/public/projects").permitAll()
                 .antMatchers("/api/logout-url").permitAll()
                 .antMatchers("/api/**")
                 .authenticated() // Allow management/health endpoint to all to allow kubernetes to be able to
--- a/src/main/java/org/radarbase/management/config/OAuth2ServerConfiguration.kt
+++ b/src/main/java/org/radarbase/management/config/OAuth2ServerConfiguration.kt
@@ -129,6 +129,7 @@ class OAuth2ServerConfiguration(
                 .skipUrlPattern(HttpMethod.GET, "/management/health")
                 .skipUrlPattern(HttpMethod.POST, "/oauth/token")
                 .skipUrlPattern(HttpMethod.GET, "/api/meta-token/*")
+                .skipUrlPattern(HttpMethod.GET, "/api/lite-projects")
                 .skipUrlPattern(HttpMethod.GET, "/api/sitesettings")
                 .skipUrlPattern(HttpMethod.GET, "/api/redirect/**")
                 .skipUrlPattern(HttpMethod.GET, "/api/logout-url")
@@ -173,6 +174,7 @@ class OAuth2ServerConfiguration(
                 .hasAnyAuthority(RoleAuthority.SYS_ADMIN_AUTHORITY)
                 .antMatchers("/api/profile-info").permitAll()
                 .antMatchers("/api/sitesettings").permitAll()
+                .antMatchers("/api/lite-projects").permitAll()
                 .antMatchers("/api/logout-url").permitAll()
                 .antMatchers("/api/**")
                 .authenticated() // Allow management/health endpoint to all to allow kubernetes to be able to
--- a/src/main/java/org/radarbase/management/service/ProjectService.kt
+++ b/src/main/java/org/radarbase/management/service/ProjectService.kt
@@ -72,7 +72,7 @@ class ProjectService(
     /**
      *Get all the projects for the public endpoint.
      *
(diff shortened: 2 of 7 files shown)


Bdegraaf1234:
Nice work @this-Aditya,

Some thoughts, feel free to override:

- We already have a class MinimumprojectDetails, is that perhaps suitable to expand here?
- We might benefit from naming these public endpoints something like `api/projects/public` and putting it in the `ProjectResource`.
- Alternatively, I would switch to projects-lite so that any alphabetically ordered endpoint listing will group this with `api/projects`

I guess I just logically group this with Project a little more than is done in the current implementation. If others disagree (@yatharthranjan @mpgxvii) feel free to approve and move forward

this-Aditya:
Thanks for the review, @Bdegraaf1234 !
I’ll make the desired modifications. Additionally, would it be acceptable to also retrieve the updated `MinimalProjectDetails` when fetching from the `api/projects` endpoint with the `minimized` parameter set to `true`?

Bdegraaf1234:
Yeah that might be an issue, I think it would be good to check with if someone knows where we use the `minimized` endpoint..

Are you in the SEP meeting later today? Would be good to discuss this there

this-Aditya:
Okay, fine. I don't have an invite for that meeting. Could you please share the link on Slack? 
Thanks.

src/main/java/org/radarbase/management/service/ProjectService.kt
```diff
@@ -69,6 +69,17 @@ class ProjectService(
         }
     }
 
+    /**
+     *Get all the projects for the public endpoint.
+     *
+     *@return the list of ProjectLiteDTO
+     */
+    @Transactional(readOnly = true)
+    fun findProjectsForPublicEndpoint(pageable: Pageable): Page<*> {
```
yatharthranjan, line 78: `getPublicProjects`

src/main/java/org/radarbase/management/config/OAuth2ServerConfiguration.kt
```diff
@@ -129,6 +129,7 @@ class OAuth2ServerConfiguration(
                 .skipUrlPattern(HttpMethod.GET, "/management/health")
                 .skipUrlPattern(HttpMethod.POST, "/oauth/token")
                 .skipUrlPattern(HttpMethod.GET, "/api/meta-token/*")
+                .skipUrlPattern(HttpMethod.GET, "/api/lite-projects")
```
yatharthranjan, line 132: I agree with Bastiaan, this should be something like /api/public/projects

src/main/java/org/radarbase/management/service/dto/ProjectLiteDTO.kt
```diff
@@ -0,0 +1,55 @@
...
+    var projectName: String? = null
+
+    @NonNull
+    var description: String? = null
+
+    @NotNull
+    var location: String? = null
+    var projectStatus: ProjectStatus? = null
+
+
```
yatharthranjan, line 21: we need source-types here

src/main/java/org/radarbase/management/service/dto/ProjectLiteDTO.kt
```diff
@@ -0,0 +1,55 @@
+package org.radarbase.management.service.dto
+
+import com.fasterxml.jackson.annotation.JsonInclude
+import org.radarbase.management.domain.enumeration.ProjectStatus
+import reactor.util.annotation.NonNull
+import java.io.Serializable
+import javax.validation.constraints.NotNull
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+class ProjectLiteDTO : Serializable {
```
yatharthranjan, line 10: `PublicProjectDTO`

src/main/java/org/radarbase/management/web/rest/ProjectLiteResource.kt
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * GET  /lite-projects : get all the projects.
+ *
+ * @return the ResponseEntity with status 200 (OK) and the list of ProjectLiteDTO
+ */
+
+@RestController
+@RequestMapping("/api")
+class ProjectLiteResource(
```
yatharthranjan, line 24: You can call this `PublicResource` with annotation `@RequestMapping("/api/public")` and then we can house all the public endpoints in the future here.

src/main/java/org/radarbase/management/web/rest/ProjectLiteResource.kt
```diff
@@ -0,0 +1,43 @@
...
+    @Autowired private val projectService: ProjectService
+) {
+
+    @GetMapping("lite-projects")
```
yatharthranjan, line 28: should be just `projects` with the above path.

yatharthranjan (commented):
I agree with most of Bastiaan's suggestions. Adding a new DTO instead of extending the previous one is acceptable, as this has a specific public use case. The previous one improved performance for the MP's home page (for logged-in users).

@Bdegraaf1234 I think the SEP meeting is cancelled today as a few of us are on annual leave.

@this-Aditya Please take a look at some comments below.

src/main/java/org/radarbase/management/web/rest/PublicResource.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    @GetMapping("projects")
+    @Timed
+    fun getProjectsInfo(
+        @PageableDefault(size = Int.MAX_VALUE) pageable: Pageable,
+    ): ResponseEntity<*> {
+        log.debug("REST request to get Projects for public endpoint")
+        val page = projectService.getPublicProjects(pageable)
+        val headers = PaginationUtil
+            .generatePaginationHttpHeaders(page, "/api/lite-projects")
```
Bdegraaf1234, line 36: Is this right?

src/main/java/org/radarbase/management/service/dto/PublicProjectDTO.kt
```diff
@@ -0,0 +1,48 @@
...
+@JsonInclude(JsonInclude.Include.NON_NULL)
+class PublicProjectDTO : Serializable {
+    @NonNull
+    var projectName: String? = null
+
+    @NonNull
+    var description: String? = null
+
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    var sourceTypes: Set<SourceTypeDTO> = emptySet()
```
yatharthranjan, line 18: I think here, MinimalSourceTypeDTO should be enough

src/main/java/org/radarbase/management/web/rest/PublicResource.kt
this-Aditya: Ah, sorry for the mistake. I've corrected it in the latest commit.

yatharthranjan (approved):
Thanks for the updates. Nice work!

LGTM

@mpgxvii do you wish to review?

yatharthranjan:
Merging for now. We can create separate PRs if changes are needed

PR #948, opened 2024-09-03
Pass url in httpclient

mpgxvii:
Description: 
- Pass url in httpclient

#### Checklist:


radar-auth/src/main/java/org/radarbase/auth/jwks/JwksTokenVerifierLoader.kt
```diff
@@ -78,7 +78,7 @@ class JwksTokenVerifierLoader(
 
         private val logger = LoggerFactory.getLogger(JwksTokenVerifierLoader::class.java)
 
-        private val httpClient = HttpClient(CIO).config {
+        private fun httpClient(url: String) = HttpClient(CIO).config {
```
yatharthranjan, line 81: you can move this to the class property instead of converting to a function. I don't think the url will change per class instance

pvannierop:
Replaced by https://github.com/RADAR-base/ManagementPortal/pull/949

PR #971, opened 2024-10-30
[bugfix] Escape project name in MP request

pvannierop:
# Problem 
Rest source backend has an error when a project name contains a space. This is caused by the management-portal client not encoding the URL before making the request.

# Solution
This PR will add escape of the URL for the api/projects/$projectId/subjects endpoint.


keyvaann (approved):
LGTM

PR #995, opened 2025-02-11
Add scheduled Snyk check for Docker image

keyvaann:
This PR adds a weekly scan by Snyk to check the Docker image and will suggest fixes if it fails.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -78,7 +78,7 @@ jobs:
 
       - name: Upload screenshots of failed e2e tests
         if: always()
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           path: src/test/javascript/e2e/cypress/screenshots
           if-no-files-found: ignore
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -71,7 +71,7 @@ jobs:
 
       - name: Upload screenshots of failed e2e tests
         if: always()
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           path: src/test/javascript/e2e/cypress/screenshots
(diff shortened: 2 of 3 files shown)


mpgxvii:
@keyvaann I think the main test workflow is failing because of the deprecated action `upload-artifact`. We can just update to v4 and the tests will run again. I've updated here: https://github.com/RADAR-base/ManagementPortal/pull/996/commits/6c775c6f2ea1e243eda18abb2526ed794ac82878

mpgxvii (dismissed):
LGTM! Just minor comment.

mpgxvii (approved):

mpgxvii (approved):
Thanks! Also, are we merging to master directly?

keyvaann:
@mpgxvii I don't know where is the best place to merge it. Maybe @pvannierop has suggestions?

pvannierop:
@mpgxvii @keyvaann 

> Thanks! Also, are we merging to master directly?

Depends on the state of the _dev_ branch. We need to make a periodic release that fixes vulnerabilities. I rather not have to deal with problems that arise from features that are in development. I, therefore, would like a PR directly to master.

pvannierop (commented):
I have two issues that I would like to raise:
- Should the scheduled action only trigger on _master_ branch?
- Should the scheduled docker action replace the current Snyk scan that is performed for each commit to a PR branch?

keyvaann:
> I have two issues that I would like to raise:
> 
> * Should the scheduled action only trigger on _master_ branch?
> * Should the scheduled docker action replace the current Snyk scan that is performed for each commit to a PR branch?

It's probably better to run it only on master branch since this is used to fix security vulnerabilities. 
I think having a snyk scan that directly checks the codebase is better than the docker image scan so I think the current snyk scan in each commit is good, but I don't know how much attention is actually being paid to it.

PR #997, opened 2025-02-12
Fixed the snyk scan CI

keyvaann:
Fixes the issue that was caused by the previous PR.


pvannierop (approved):
Thank you. Let's see whether this solves the problem.

PR #1087, opened 2026-09-24
Fix login with internal auth server when public URL is not reachable in-pod

pvannierop:
## Problem

With `managementportal.authServer.internal=true`, login redirects straight back to the login page, and nothing is logged. After a successful form login, `/api/redirect/login?code=...` exchanges the code at `<managementPortalBaseUrl>/oauth/token`. That is the **public** URL, and it isn't always reachable from inside the container. On a local k3d deployment, `http://localhost/managementportal` resolves to the pod itself on port 80, where nothing listens. The resulting `IdpException` was caught without logging, and the user was sent to `/error?message=Unable%20to%20authenticate`.

`InternalTokenValidatorConfiguration` has the same problem for `<managementPortalBaseUrl>/oauth/token_key`.

## Fix

- Add `InternalServerUrl`, which gives `http://localhost:<server.port><server.servlet.context-path>`.
- Use it for the self-calls that happen only in internal auth server mode: the code exchange in `LoginService` and the `token_key` JWKS loader in `InternalTokenValidatorConfiguration`.
- External auth server mode (`authServer.internal=false`, Hydra) is unchanged: it still uses `authServer.tokenUrl` / `authServer.jwksUrl`.
- Log the exception in `LoginEndpoint` when login or token validation fails.

## Testing

- `./gradlew compileKotlin` passes.
- Local k3d cluster (RADAR-Kubernetes, `server_name: localhost`, `authServer.internal=true`):
  - Before: form login succeeded, then `/api/redirect/login?code=...` redirected to `/error?message=Unable%20to%20authenticate`, and nothing was logged.
  - After (image built from this branch): `/api/redirect/login?code=...` redirects to `/managementportal/#/`, and `/api/account` returns the admin account.
  - In this setup Hydra isn't deployed, so MP logs a warning that it can't fetch the Hydra JWKS (`authServer.jwksUrl`). Validation still succeeds with the internal `token_key` loader.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mpgxvii (approved):
LGTM! Thanks for the fix.

